PR #1, opened 2016-11-02
GPII-1596

michelled:
I'm hoping to get some feedback on how I Infusionized the wrapper. 

This is not ready to go in for several reasons: 
1. there are no tests for the actual wrapper
2. the existing unit tests do not pass due to https://issues.gpii.net/browse/GPII-2091
3. the snapsets need to be hidden behind a demo or debug flag


tests.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+
+const {app} = require('electron')
```
amb26: What is this strange syntactical jargon : P

package.json
```diff
@@ -0,0 +1,31 @@
...
+    "tutorial"
+  ],
+  "devDependencies": {
+    "electron": "1.4.1",
+    "request": "^2.69.0",
+    "shelljs": "0.3.0"
+  },
+  "dependencies": {
+    "infusion": "git://github.com/fluid-project/infusion.git#99765614ca5679580024e8b673189ed419468a0c",
+    "windows": "git://github.com/javihernandez/windows.git#6a1d8fdb0836c6ff3aa9b355fa83ae8938680558",
```
amb26: Are there particular fixes you require from these branches? It would be helpful to add a README explaining which branch they are bound to and why we need them

package.json
```diff
@@ -0,0 +1,31 @@
...
+    "start",
+    "tutorial"
+  ],
+  "devDependencies": {
+    "electron": "1.4.1",
+    "request": "^2.69.0",
+    "shelljs": "0.3.0"
+  },
+  "dependencies": {
+    "infusion": "git://github.com/fluid-project/infusion.git#99765614ca5679580024e8b673189ed419468a0c",
```
amb26: Bind to a dev release if you can, so we can get at least some insight into which version of Infusion this is

tests.js
```diff
@@ -0,0 +1,7 @@
...
+
+app.on("ready", function() {
+    require("./node_modules/gpii-windows/tests/UnitTests.js");
```
amb26: This hard-coding to a path relative to node_modules will be unstable between versions of npm.
Instead, I suggest something like
var fluid = require("infusion");
require("gpii-windows");
fluid.require("%gpii-windows/tests/UnitTests.js");

main.js
```diff
@@ -0,0 +1,181 @@
...
+    request("http://localhost:8081/user/"+token+"/logout", function(error, response, body) {
+        //TODO Put in some error logging
+    });
+}
+
+
+// Component to create and update the task tray menu.
+fluid.defaults("gpii.taskTray", {
+    gradeNames: "fluid.modelComponent",
+    model: { // Do I need to specify a default or will it sync up with the other component?
```
amb26: Which other component? There's no connection between the models of different components unless you set one up with a Model Relay definition: http://docs.fluidproject.org/infusion/development/ModelRelay.html

Perhaps what you want to write here is model: "{that}.app.model" ?
In which direction do you want the syncing to travel?

main.js
```diff
@@ -0,0 +1,181 @@
...
+    listeners: {
+        onCreate: [{
+            funcName: "gpii.app.startLocalFlowManager"
+        }, {
+            funcName: "{that}.changeStarted",
+            args: [true]
+        }]
+    },
+    modelListeners: {
+        keyedInSet: [{ //  Will this be a race condition?
```
amb26: Yes, it probably will. Please put priority-based constraints on these if you want them to execute in a particular order:

http://docs.fluidproject.org/infusion/development/Priorities.html#table-of-supported-sites-for-priorities-and-namespaces

main.js
```diff
@@ -0,0 +1,181 @@
...
+        changeSet: {
+            changePath: "keyedInSet",
+            value: "{arguments}.0"
+        }
+    },
+    icon: "web/icons/gpii.ico"
+});
+
+gpii.taskTray.makeTray = function (icon) {
+    tray = new Tray(path.join(__dirname, icon));
```
amb26: Tough that this object reference will be lost forever. Attach it to the component using an expander inside a member definition instead: https://github.com/fluid-project/infusion/blob/master/src/components/reorderer/js/Reorderer.js#L202

main.js
```diff
@@ -0,0 +1,181 @@
...
+    } else {
+        menu.push({ label: "Not keyed in", enabled: false });
+    }
+    return menu;
+};
+
+gpii.taskTray.updateSnapsets = function (menu, keyInFn) {
+    menu.push({ label: "Key in set...",
+        submenu: [
+            { label: "Alice", click: function() { keyInFn("alice"); }},
```
amb26: Build these automatically from a block in the component's options

main.js
```diff
@@ -0,0 +1,181 @@
...
+});
+
+gpii.taskTray.makeTray = function (icon) {
+    tray = new Tray(path.join(__dirname, icon));
+    tray.setToolTip("GPII Electron");
+};
+
+gpii.taskTray.updateSet = function (menu, setName, changeSetFn) {
+    if (setName) {
+        menu.push({ label: "Keyed in with " + setName, enabled: false });
```
amb26: Put these strings into a block of options, perhaps named "menuLabels" or so

main.js
michelled: Thanks! Addressed at 0bce4d2eff0ef9270f782792efbf4481bf4c7b43

main.js
michelled: Thanks! Addressed at e03b99066a65fafec0efbf79a53d084529910747

main.js
michelled: Addressed at c024cb43a14a5d1876837f8a56927f9f02ff5118

main.js
michelled: Addressed at 0bce4d2eff0ef9270f782792efbf4481bf4c7b43

package.json
```diff
@@ -0,0 +1,31 @@
...
+    "url": "git+https://github.com/gpii/gpii-electron.git"
+  },
+  "keywords": [
+    "Electron",
+    "quick",
+    "start",
+    "tutorial"
+  ],
+  "devDependencies": {
+    "electron": "1.4.1",
```
michelled: Added at 264b12d6546b706db740d2a668e9aae833a603b3

tests.js
michelled: Fixed at 264b12d6546b706db740d2a668e9aae833a603b3

package.json
michelled: No, I just needed the Node upgrade. I've switched to the master branches again.

main.js
michelled: Yes! That was what I wanted. :) 
Fixed at ac8300c20e9da19d84f7d8b0368e521e8bea0a42

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "eslint-config-fluid": "^1.0.0",
+    "fluid-grunt-eslint": "18.1.2",
+    "grunt": "1.0.1",
+    "grunt-jsonlint": "1.1.0",
+    "grunt-shell": "1.3.0",
+    "request": "^2.69.0",
```
amb26: This and electron are in the wrong block since they are part of the implementation. Also, our policy is to use exact versions throughout.

src/app.js
```diff
@@ -0,0 +1,247 @@
+/*!
+GPII Electron
```
amb26: Stale comment

src/app.js
```diff
@@ -0,0 +1,247 @@
...
+        label: exitLabel,
+        click: function () {
+            var app = require("electron").app;
+            app.quit();
+        }
+    });
+    return menu;
+};
+
+gpii.taskTray.menu.updateMenu = function (that, gpiiStarted, keyedInSet) {
```
amb26: Doc comments required for these functions, as seen in modern code such as https://github.com/GPII/universal/blob/master/gpii/node_modules/gpii-oauth2/gpii-oauth2-authz-server/src/AuthorizationService.js#L183

src/app.js
```diff
@@ -0,0 +1,247 @@
...
+    });
+    return menu;
+};
+
+gpii.taskTray.addExit = function (menu, exitLabel) {
+    menu.push({
+        label: exitLabel,
+        click: function () {
+            var app = require("electron").app;
+            app.quit();
```
amb26: Recall earlier feedback that this needs to fire an event which may cause a quit, rather than be hardwired in the menu structure. Without this, your code will be impossible to test. In general, the hard-wired click functions are a bad idea. I suggest that instead you provide each of them with a standard "click relay" which rather than including an implementation, just forward to a generic "menuItemClicked" event with appropriate arguments - or at the very least some component event specific to the menu item.

src/app.js
```diff
@@ -0,0 +1,247 @@
...
+        davey: "Davey",
+        david: "David",
+        elaine: "Elaine",
+        elmer: "Elmer",
+        elod: "Elod",
+        livia: "Livia"
+    }
+});
+
+gpii.taskTray.updateSet = function (menu, menuLabels, setName, changeSetFn) {
```
amb26: Peculiarly named function with a peculiar final argument. Perhaps something like "renderSnapsetMenu" or so? Also see later comments about degearing from raw function handles and handling menu clicks a little more generically, using events.

src/app.js
```diff
@@ -0,0 +1,247 @@
...
+        //TODO Put in some error logging
+    });
+};
+
+/*
+ ** Component to create and update the task tray menu.
+ */
+fluid.defaults("gpii.taskTray.menu", {
+    gradeNames: "fluid.modelComponent",
+    menuLabels: {
```
amb26: Try to declaratively encode the effects of the menu items as well as just the labels. It will give your users a central point of control and clean up your code considerably.

src/app.js
```diff
@@ -0,0 +1,247 @@
...
+    var fluid = require("universal"),
+        gpii = fluid.registerNamespace("gpii");
+
+    if (os.platform() === "win32") {
+        require("gpii-windows/index.js");
+    }
+
+    gpii.start();
+    //TODO: needs to fire an on ready so we can be certain that everything has started
+    that.events.onAppReady.fire();
```
amb26: It is a little awkward, but there is a certain entitlement that an express server may actually start asynchronously (although I have not yet observed this in practice). To be absolutely secure that the GPII has really started, you will need to inject down a listener to this event in the Kettle server: https://github.com/fluid-project/kettle/blob/master/lib/KettleServer.js#L68
If we are short of time this can be left as a TODO since it will require a sound understanding of the geometry of everything as well as a bit of options distribution 
http://docs.fluidproject.org/infusion/development/IoCSS.html#-distributeoptions-and-iocss-downward-matching-css-like-context-selectors-for-options-forwarding

src/app.js
```diff
@@ -0,0 +1,247 @@
...
+            args: ["{that}"]
+        },
+        "onCreate.changeStarted": {
+            funcName: "{that}.changeStarted",
+            args: [true]
+        }
+    },
+    modelListeners: {
+        keyedInSet: [{
+            funcName: "gpii.app.keyOut",
```
amb26: Stick some namespaces on these so you can find them again

src/app.js
```diff
@@ -0,0 +1,247 @@
...
+Copyright 2016 OCAD University
+
+Licensed under the New BSD license. You may not use this file except in
+compliance with this License.
+The research leading to these results has received funding from the European Union's
+Seventh Framework Programme (FP7/2007-2013) under grant agreement no. 289016.
+You may obtain a copy of the License at
+https://github.com/GPII/universal/blob/master/LICENSE.txt
+*/
+/* eslint-env node */
```
amb26: This directive can be placed in your eslintrc since all of your files are node-only

tests/testMain.js
```diff
@@ -0,0 +1,56 @@
...
+
+var fluid = require("infusion");
+var gpii = fluid.registerNamespace("gpii");
+var jqUnit = fluid.require("node-jqunit");
+
+require("../src/app.js");
+
+jqUnit.module("GPII Application Main Tests");
+
+// jqUnit.asyncTest("Testing App", function () {
```
amb26: Actual tests to appear :)

src/app.js
michelled: Addressed at 56bccc7632751c202ccff99edc01bb113222d5d1

src/app.js
michelled: Addressed at 56bccc7632751c202ccff99edc01bb113222d5d1

src/app.js
michelled: Addressed at ecb5598b97a4c853c220724c3207a9cab7c2c02e

src/app.js
michelled: Addressed at bc3e03d8c22c044b012e2419897dbbd2de0b56ff

package.json
michelled: Addressed at bc3e03d8c22c044b012e2419897dbbd2de0b56ff

src/app.js
michelled: Does this implement the "click relay" you were envisioning? 
https://github.com/michelled/gpii-app/blob/bc3e03d8c22c044b012e2419897dbbd2de0b56ff/src/app.js#L249

src/app.js
```diff
@@ -0,0 +1,306 @@
...
+            },
+            createOnEvent: "onGPIIReady"
+        }
+    },
+    components: {
+        menu: {
+            type: "gpii.app.menu",
+            createOnEvent: "onGPIIReady",
+            options: {
+                model: {
```
amb26: All of this material should be factored out into a top-level addin grade named, say, gpii.app.menuInApp

src/app.js
michelled: This has been addressed through refactoring the menu component. cc504d443bd10954a9c946f753d844a4d7556f32

src/app.js
michelled: This has been addressed through refactoring the menu component. cc504d4

configs/app.dev.json
```diff
@@ -0,0 +1,19 @@
+{
```
javihernandez: @michelled thanks!
I guess that we should include into the README file the instructions to run the app with a specific config.

michelled: I'm actually going to change that so it's easier as per @cindyli's suggestion in the channel yesterday. But I figured I might as well get you what you need for HST first. :)

configs/app.dev.json
```diff
@@ -0,0 +1,19 @@
...
+    "type": "gpii.appWithTaskTray",
```
cindyli: Please distinguish the type name in each config file to be more specific with the goal of this config.

cindyli: As this config is based off the other config `app.json`. It can be simplified to:
```
{
    "type": "gpii.appWithTaskTray.dev",
    "options": {
        "gradeNames": ["fluid.component"],
            "distributeMenuDisplay": {
                "record": "gpii.app.menuInAppDev",
                "target": "{that menu}.options.gradeNames"
            }
        }
    },
    "mergeConfigs": [
        "./app.json"
     ]
}
```

src/app.js
michelled: Addressed at 448f1c1b894ec7c6d1860e29163ec67b2944e86e

PR #4, opened 2017-02-13
GPII-1596: Removed the snapsets from the menu in production mode. Add…

michelled:
…ed a dev mode config file.


michelled:
@javihernandez this pulls out the snapsets from the task tray menu in the default production mode. There is a dev config file that can be used when we want quick access to snapsets.

PR #5, opened 2017-02-14
GPII-2200: Moved debug logging to after the fluid.log redirect has been set

stegru:
The debug logging was being written to stdout before fluid.doLog was re-written.

Moving the require statements for logging.js before gpii-windows didn't work because getting the settings directory in logging.js depends on the windows context.


stegru:
Oh, this loosely depends on a change to windows, where the immediate logging has been stopped. Without it, the debug info will be written twice - before the redirect and then after.

PR #6, opened 2017-02-14
GPII-2282: Logout any logged in user before closing the application

javihernandez:


javihernandez:
@michelled, does this make sense to you? If so, since I'm the only person merging things into hst-2017, who can merge this in?

src/app.js
```diff
@@ -127,18 +129,37 @@ gpii.app.keyIn = function (token) {
...
     var app = require("electron").app;
     app.quit();
+}
+
+/**
+  * Handles the exit of the Electron Application.
+  * @param that {Component} An instance of gpii.app
+  */
+gpii.app.exit = function (that) {
+    //TODO: This should stop the GPII gracefully before quitting the application.
```
amb26: I'd say this is sufficient, the GPII does not hold any particular state when the user is not logged in. If you like, you could use @sgithens brand new GPII DOT STOP method: https://github.com/GPII/universal/blob/master/index.js#L91

javihernandez: done

src/app.js
```diff
@@ -127,18 +129,37 @@ gpii.app.keyIn = function (token) {
   * @param token {String} The token to key out with.
   */
 gpii.app.keyOut = function (token) {
```
amb26: Need to update the docs of this function so that it is clear that it initiates the keyOut action immediately and returns a promise for its completion.

src/app.js
```diff
@@ -127,18 +129,37 @@ gpii.app.keyIn = function (token) {
...
     app.quit();
+}
+
+/**
+  * Handles the exit of the Electron Application.
+  * @param that {Component} An instance of gpii.app
+  */
+gpii.app.exit = function (that) {
+    if (that.model.keyedInUserToken) {
+        fluid.promise.map(that.keyOut(that.model.keyedInUserToken), gpii.app.performQuit);
```
amb26: This is a misleading use of fluid.promise.map since performQuit makes no use of its argument and has no return value. I suggest upgrading gpii.app.performQuit to a (synchronous) promise-returning function and the use of fluid.promise.sequence instead

javihernandez:
I've just updated it according to review comments.

src/app.js
```diff
@@ -125,20 +127,46 @@ gpii.app.keyIn = function (token) {
...
+    return togo;
+}
+
+/**
+  * Handles the exit of the Electron Application.
+  * @param that {Component} An instance of gpii.app
+  */
+gpii.app.exit = function (that) {
+    if (that.model.keyedInUserToken) {
+        fluid.promise.sequence([that.keyOut(that.model.keyedInUserToken), gpii.app.performQuit]);
```
amb26: Now you have a failure handler in keyOut, it is now possible that failure of the keyout request can abort the quit action. You should either wrap the the request promise using a utility such as this one, https://github.com/GPII/universal/blob/master/gpii/node_modules/lifecycleManager/src/LifecycleManager.js#L353 , or else put similar code inside keyOut itself.

javihernandez: done

PR #9, opened 2017-02-24
GPII-2290: Clone gpii-hst-tools to include it into the installer

javihernandez:


provisioning/Installer.ps1
```diff
@@ -29,6 +29,13 @@ if ((Test-Path -Path $installerDir)){
 }
 Invoke-Command $git "clone --branch $($installerBranch) $($installerRepo) $($installerDir)"
 
+# If gpii-hst-tools exists delete it and clone it again
+$hstToolsDir = Join-Path (Join-Path $installerDir "staging") "gpii-hst-tools"
+if ((Test-Path -Path $installerDir)){
```
stegru: It should be checking if $hstToolsDir exists instead?
And are the double brackets intentional?

javihernandez: My fault, copy/paste side effect :/
Good catch @stegru!! I've just updated it and removed the double brackets since they aren't needed.

PR #10, opened 2017-03-14
GPII-2312: Updated gpii-app icons

javihernandez:
This is a first iteration of the icon for the gpii-app, it works well on Windows [10](https://issues.gpii.net/secure/attachment/15302/new_icon_windows10.png), [8.1](https://issues.gpii.net/secure/attachment/15300/new_icon_windows81.png) and [7](https://issues.gpii.net/secure/attachment/15301/new_icon_windows7.png).


amatas:
ok to test

gtirloni:
ok to test

colinbdclark:
The icon is still a tiny bit fuzzy and jaggy, but I think given the constraints, we should just ship it. Just for posterity, @javihernandez, can you summarize here a few of the issues you encountered when dealing with Windows icons and Electron?

javihernandez:
@colinbdclark, yes, although this was supposed to be an easy task , I've found a strange behavior when trying to create an icon that works well in the task tray. The main problem comes from using the .ico file. After exporting the original svg file into an .ico file (with different sizes, 32x32, 64x64, etc) the icon gets blurry when inserted into the system tray. After doing some tests, I could check that Windows is not taking the right size of the icon but taking another one (apparently, it takes the biggest one) and applying its own scaling to fit into the task tray, and this makes the icon to look blurry. I could ensure this behavior by providing an .ico file that only includes the 32x32 size icon, which looks as expected. This behavior has been reported in this pull request (https://github.com/electron/electron/pull/5624#issuecomment-221932353), so looks like it's a problem in electron, which is not passing the right icon to Windows. Hence the reason why Windows scales the icon to make it fit into the task tray.

javihernandez:
@kaspermarkus do you mind taking care of this? We already have a +1 from @colinbdclark

PR #17, opened 2017-07-19
Gpii 2092

michelled:
@amb26 this is a pull request to get some guidance on completing these tests. The questions are in the code.

Specifically, when I try to key in a person while someone is already keyed in, the request should throw an error. The error response I’m interested in is being set here: https://github.com/GPII/universal/blob/master/gpii/node_modules/flowManager/src/UserLogonStateChange.js#L283
I try to listen to it here: https://github.com/michelled/gpii-app/blob/GPII-2092/tests/IntegrationTestDefs.js#L84 but it just hangs there.

Secondly, I would like to test the app exit but the tests return without printing the results. 
https://github.com/michelled/gpii-app/blob/GPII-2092/tests/IntegrationTestDefs.js#L93

Finally, I want to be able to test the dev config as well. How do I stop the server and restart it with a different config? 
https://github.com/michelled/gpii-app/blob/GPII-2092/tests/IntegrationTestDefs.js#L114

Thanks for your help!


amb26:
i) Yes, that is a terrible way of listening to the request error. IoCSS was designed just to avoid such brittle, hard-coded paths through the tree. In fact, "request" is a dynamicComponent of requests, so its actual member name will increment on every request. Also, it may well not be allocated at the time you begin to listen to the request. http://docs.fluidproject.org/infusion/development/SubcomponentDeclaration.html#naming-of-dynamic-components

The best way to hit the request you are interested in is with an IoCSS binding for example like {that flowManager kettle.request.http}.events.onError - remember that there is a special dispensation in the IoC Testing framework to bind to events using IoCSS - http://docs.fluidproject.org/infusion/development/IoCTestingFramework.html#supported-fixture-records

ii) You should hoist a reference to the menu out to the top level of your testCaseHolder to avoid brittle paths like {configuration}.server.flowManager.app.tray.menu.applier.modelChanged - you can use a hack like the one we use in the GPII's own testing framework, as https://github.com/GPII/universal/blob/master/gpii/node_modules/testing/src/Testing.js#L254

iii) To test another config, you'll just need to write a fresh set of fixtures. That is, write a new set of testDefs with a different value for the config https://github.com/michelled/gpii-app/blob/GPII-2092/tests/IntegrationTestDefs.js#L63 (some confusion here given this seems to be configName rather than the configType we look for here https://github.com/michelled/gpii-app/blob/GPII-2092/tests/IntegrationTests.js#L67 - should look at what is going on here), and then send an array of them in to  gpii.tests.app.bootstrapServer on https://github.com/michelled/gpii-app/blob/GPII-2092/tests/IntegrationTests.js#L87

amb26:
iv) yes, your current exit implementation performs a real exit of the app. You'll need to override this invoker https://github.com/michelled/gpii-app/blob/GPII-2092/src/app.js#L91 with some kind of mock which perhaps does nothing or fires an event you are interested in.

gtirloni:
@amb26 @michelled when the tests get stuck, the same URL is requested over and over (https://preferences.gpii.net/preferences/sammy) and nothing else happens.

 * Plenty of idle CPU
 * Plenty of free memory
 * Barely any disk activity
 * electron.exe process does absolutely nothing but open a connection to preferences.gpii.net and close it successfully every few minutes
 * Requests to preferences.gpii.net inside and outside the VM are averaging 900ms for me (huge delay due to my location but still doesn't explain >1min delays between log lines -- is it probably a timer for the heartbeat check that keeps working when other things are stuck?)

This log pattern repeats itself over and over forever:

```
14:51:00.060:  contextManager: Updating context with: {
    "http://registry.gpii.net/common/environment/timestamp": 1503525060060,
    "http://registry.gpii.net/common/environment/timeOfDay": "14:51"
}
14:51:20.068:  contextManager: Updating context with: {
    "http://registry.gpii.net/common/environment/timestamp": 1503525080068,
    "http://registry.gpii.net/common/environment/timeOfDay": "14:51"
}
14:51:38.476:  DataSource Issuing HTTPS request with options {
    "port": null,
    "method": "GET",
    "headers": {

    },
    "protocol": "https:",
    "auth": null,
    "host": "preferences.gpii.net",
    "hostname": "preferences.gpii.net",
    "path": "/preferences/sammy",
    "url": "https://preferences.gpii.net/preferences/sammy",
    "termMap": {

    },
    "directModel": null,
    "operation": "get",
    "reverse": false,
    "writeMethod": "PUT"
}
14:51:38.555:  contextManager: Updating context with: {
    "http://registry.gpii.net/common/environment/timestamp": 1503525098555,
    "http://registry.gpii.net/common/environment/timeOfDay": "14:51"
}
14:51:39.127:  Network diagnostic request to url https://preferences.gpii.net/preferences/sammy successful with response
```

Trying to isolate the issue...

 * Running only AppTests.js: `15/15 total passed`
 * Running only IntegrationTests.js: the `failure in fixture file - no tests were queued` error is reported. However, it continues running and finishes with success (14/14 total passed). The exit code is zero (success).
 * Running App + Integration tests consecutively, same combined results (final result is success, nothing stuck).
 * Running only `universal/tests/all-tests`: `1001/1001 total passed`
 * Running just `gpii-windows/tests/UnitTests`: `118/121 total passed`.
* Running `universal/tests/all-tests` following by `gpii-windows/tests/UnitTests`: `1126/1127 total passed`

I'm not sure if these tests can be execute over and over on the same environment (because they seem to change things), so please excuse my ignorance if these tests are meaningless.

amb26:
The diagnostic request is just something I inserted into the HST branch for https://issues.gpii.net/browse/GPII-2349 and isn't relevant to the problem.

There is some kind of timing issue with the tests overall that is either causing premature exit of the process or preventing timely exit of the process. My initial guess is that this is some further kind of timing disturbance, of the kind we've seen before, where perhaps the greater startup time of an electron process is interfering with the normally perfectly safe debounce timeout that is used in order to time process exit in node-jqUnit here: https://github.com/fluid-project/node-jqunit/blob/master/lib/jqUnit-node.js#L181

The fact that @michelled was able to disturb the symptoms of the problem by messing with the near-duplicate of this that we have here: https://github.com/michelled/gpii-app/blob/GPII-2092/tests.js#L15 also supports the idea that the problem is in this area. We can get some greater insight into the problem by monkey-patching these sites in order to provide some extra logging indicating when the system has come to believe that its queue is drained of tests or not.

michelled:
@amb26 more review on this branch, please :)

michelled:
ok to test

michelled:
ok to test

amb26:
@javihernandez - note that the test failure is within the hst-2017 branch of gpii-windows itself - I can reproduce it by running its unit tests. The failure is also in Windows master - it looks like your merge of @klown 's changes at https://github.com/GPII/windows/commit/fbcb422823690144ce2cba802722a9ef7cc0b576 is bad

amb26:
The trailing failures I noted above are still present - the line numbers have changed a little:

```
04:00:13.618:  FATAL ERROR: Uncaught exception: Object has been destroyed
Error: Object has been destroyed
    at Error (native)
    at Timeout._repeat (C:\gpii-app\src\app.js:429:26)
    at Timeout.wrapper [as _onTimeout] (timers.js:425:11)
    at tryOnTimeout (timers.js:232:11)
    at Timer.listOnTimeout (timers.js:202:5)
```

Please do proper cleanup of this timer

klown:
> Please do proper cleanup of this timer

Shameless plug:  this is the kind of thing that the [StateTracker component](https://github.com/klown/infusion/blob/FLUID-6090/src/components/stateTracker/js/StateTracker.js) is for.

But, too much work to incorporate it right now.

src/app.js
```diff
@@ -335,6 +276,192 @@ fluid.onUncaughtException.addListener(function (err) {
...
+ * the window, the function waits until `dialogMinDisplayTime` has passed before dismissing it.
+ *
+ * @param that {Object} the app
+ */
+gpii.app.dismissWaitDialog = function (that) {
+    // ensure we have displayed for a minimum amount of `dialogMinDisplayTime` secs to avoid confusing flickering
+    var remainingDisplayTime = (that.model.dialogStartTime + that.model.dialogMinDisplayTime) - Date.now();
+
+    if (remainingDisplayTime > 0) {
+        that.model.timeout = setTimeout(function () {
```
amb26: One musn't write directly into a model by assignment - this kind of free-form material belongs at top level on the component or as a "member". Also, please give this a more descriptive name than "timeout". In fact the leak causing the failure isn't at this site but above it on line 436, but it is good to clean up this leak as well.

src/app.js
```diff
@@ -335,6 +276,192 @@ fluid.onUncaughtException.addListener(function (err) {
...
+ * @param that {Object} the app module
+ */
+gpii.app.displayWaitDialog = function (that) {
+    that.dialog.show();
+    // Hack to ensure it stays on top, even as the GPII autoconfiguration starts applications, etc., that might
+    // otherwise want to be on top
+    // see amongst other: https://blogs.msdn.microsoft.com/oldnewthing/20110310-00/?p=11253/
+    // and https://github.com/electron/electron/issues/2097
+    var interval = setInterval(function () {
+        if (!that.dialog.isVisible()) {
```
amb26: The leak causing the failure is here

amb26:
Closed for #26 which is now merged

PR #24, opened 2017-09-01
Full update of gpii-app with the last versions of both windows and universal

javihernandez:
Also, this includes:
* updated refs of infusion and node-jqunit
* GPII-2084: Added electron-edge dependency
* GPII-2537: Build the proximity listener before creating installer

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "electron": "1.4.1",
     "request": "2.69.0",
     "infusion": "3.0.0-dev.20170830T182157Z.392b2f8",
-    "gpii-windows": "git://github.com/javihernandez/windows.git#hst-2017+master20170901",
+    "gpii-windows": "git://github.com/GPII/windows.git#hst-2017",
     "node-jqunit": "1.1.7",
     "electron-edge": "6.5.4"
   }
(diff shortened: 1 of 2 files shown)


javihernandez:
ok to test

package.json
```diff
@@ -20,13 +20,14 @@
   },
   "scripts": {
     "start": "electron main.js",
-    "test": "powershell provisioning/Tests.ps1 -originalBuildScriptPath './node_modules/gpii-windows/provisioning'; electron tests.js"
+    "test": "powershell provisioning/Tests.ps1 -originalBuildScriptPath './node_modules/gpii-windows/provisioning'; electron tests-universal.js; electron tests-windows.js"
```
michelled: By separating this into two calls, we wouldn't have a consolidated number of passing tests. Are we ok with that?

javihernandez: Well, we aren't ok but I'd say that we can make the effort unless we can ensure that the tests run sequentially.  :smile: The underlying problem is the way we launch the tests, which I guess will change after https://github.com/GPII/gpii-app/pull/17 is ready, isn't it? @michelled @amb26 what do you think?

stegru: It would be better if I fixed the metrics tests to work.

michelled: @javihernandez I think we still expect to launch the tests in the same way. Also, I fear that splitting this into two calls will hide failures.

javihernandez:
ok to test

javihernandez:
ok to test

PR #27, opened 2017-10-09
PCP - first phase

danailbd:
**Work In Progress**

PR for the **first phase of the PCP**. A single PR is made as it was agreed, in order to save extensive overhead.
#### Included tasks:
- [x] [2607](https://issues.gpii.net/browse/GPII-2607)
- [x] [2610](https://issues.gpii.net/browse/GPII-2610)
- [x] [2628](https://issues.gpii.net/browse/GPII-2628)
- [x] [2636](https://issues.gpii.net/browse/GPII-2636)
- [x] [2527](https://issues.gpii.net/browse/GPII-2527)
- [x] [2528](https://issues.gpii.net/browse/GPII-2528)
- [x] [2531](https://issues.gpii.net/browse/GPII-2531)
- [x] [2532](https://issues.gpii.net/browse/GPII-2532)
- [x] [2533](https://issues.gpii.net/browse/GPII-2533)
- [x] [2593](https://issues.gpii.net/browse/GPII-2593)
- [x] [2603](https://issues.gpii.net/browse/GPII-2603)
- [x] [2608](https://issues.gpii.net/browse/GPII-2608)
- [x] [2609](https://issues.gpii.net/browse/GPII-2609)
- [x] [2634](https://issues.gpii.net/browse/GPII-2634)
- [x] [2646](https://issues.gpii.net/browse/GPII-2646)
- [x] [2657](https://issues.gpii.net/browse/GPII-2657)
- [x] [2659](https://issues.gpii.net/browse/GPII-2659)


amb26:
ok to test

src/app.js
```diff
@@ -38,24 +41,140 @@ gpii.app.electronAppListener = function () {
...
+fluid.defaults("gpii.app.gpiiConnector", {
+    gradeNames: "fluid.modelComponent",
+
+    // Configuration regarding the socket connection
+    config: {
+        gpiiWSUrl: "ws://localhost:8081/pcpChannel"
+    },
+
+    model: {
+        socket: "@expand:gpii.app.createGPIIConnection({that}.options.config)"
```
amb26: This isn't a reasonable thing to put in a model since it is not serialisable - more suitable to put in "members"

src/app.js
```diff
@@ -38,24 +41,140 @@ gpii.app.electronAppListener = function () {
...
+        showDialog: false,
+        preferences: {
+            sets: [],
+            activeSet: null
+        }
     },
     components: {
+        gpiiConnector: {
+            type: "gpii.app.gpiiConnector",
+            createOnEvent: "onPrequisitesReady",
```
amb26: Spelling: Prerequisites

src/app.js
```diff
@@ -38,24 +41,140 @@ gpii.app.electronAppListener = function () {
...
+    config: {
+        gpiiWSUrl: "ws://localhost:8081/pcpChannel"
+    },
+
+    model: {
+        socket: "@expand:gpii.app.createGPIIConnection({that}.options.config)"
+    },
+
+    events: {
+        onPreferencesUpdate: null
```
amb26: With the "on" convention it's more normal for past tense - i.e. "onPreferencesUpdated"

src/app.js
```diff
@@ -204,14 +327,223 @@ gpii.app.performQuit = function () {
...
+    var schema = settingDescriptor.schema;
+
+    return {
+        path: key,
+        value: settingDescriptor.value,
+        solutionName: settingDescriptor.solutionName,
+
+        title: schema.title,
+        description: schema.description,
+        icon: "../icons/gear-cloud-black.png",
```
amb26: Hoist this out into a component option

amb26: Sorry, ignore this comment, I see this is a placeholder for upcoming functionality - just add a TODO comment here to clarify this

src/app.js
```diff
@@ -664,36 +1141,104 @@ gpii.app.menu.getUserName = function (userToken) {
...
 
 /**
   * Generates an object that represents the menu item for keying out.
   * @param keyedInUserToken {String} The user token that is currently keyed in.
-  * @param name {String} The name of the user that is currently keyed in.
-  * @param keyOutStrTemp {String} The string to be displayed for the key out menu item.
+  * @param keyOutStr {String} The string to be displayed for the key out menu item
+  * if there is a keyed in user.
+  * @param notKeyedInStr {String} The string to be displayed when a user is not
+  * keyed in.
```
amb26: Add an @return to clarify returned structure

src/app.js
```diff
@@ -204,14 +327,223 @@ gpii.app.performQuit = function () {
...
+ * @param pcp {Object} A `gpii.app.pcp` instance
+ * @param gpiiConnector {Object} A `gpii.app.gpiiConnector` instance
+ */
+gpii.app.initPCPWindowIPC = function (app, pcp, gpiiConnector) {
+    ipcMain.on("closePCP", function () {
+        pcp.hide();
+    });
+
+    ipcMain.on("keyOut", function () {
+        console.log('HERE');
```
amb26: Stray log

src/app.js
```diff
@@ -625,30 +1078,54 @@ fluid.defaults("gpii.app.menu", {
...
             },
             priority: "last"
         }
     },
     menuLabels: {
+        pcp: "Open PCP",
         keyedIn: "Keyed in with %userTokenName",    // string template
-        keyOut: "Key out %userTokenName",           // string template
-        notKeyedIn: "Not keyed in",
+        keyOut: "Key-out of GPII",
```
amb26: Wireframes have this as "Key out of GPII"

src/app.js
```diff
@@ -38,24 +41,140 @@ gpii.app.electronAppListener = function () {
 require("electron").app.on("ready", gpii.app.electronAppListener);
 
 
+/**
+ * Responsible for creation and housekeeping of the connection to the PCP Channel WebSocket
+ */
+fluid.defaults("gpii.app.gpiiConnector", {
+    gradeNames: "fluid.modelComponent",
```
amb26: Can be just a fluid.component

src/app.js
```diff
@@ -204,14 +327,223 @@ gpii.app.performQuit = function () {
...
+
+        // Make the app feel more native
+        backgroundColor: "#71c5ef"
+    });
+
+    pcpWindow.on("blur", function () {
+        pcpWindow.hide();
+    });
+
+    var url = fluid.stringTemplate("file://%dirName/pcp/index.html", {
```
amb26: i) hoist this out into configuration
ii) expand against fluid.module.terms() rather than dirName as in https://github.com/fluid-project/kettle/blob/master/lib/dataSource-file.js#L105 so that this can read file://%gpii-app/pcp/index.html"

src/pcp/index.html
```diff
@@ -0,0 +1,44 @@
...
+        <script type="text/javascript" src="js/exemplars.js"></script>
+        <script type="text/javascript" src="js/settingsPanel.js"></script>
+        <script type="text/javascript" src="js/index.js"></script>
+    </head>
+    <body id="flc-body">
+        <div class="flc-splash">
+            <div class="logo">
+                <p><i>Not keyed in</i></p>
+                <img src="../icons/gpii-icon-balloon.png">
+                <h3>GPII - Personal Control Panel</h3>
```
amb26: Since this is user-visible it should now read "Personal Settings Panel"

src/pcp/js/exemplars.js
```diff
@@ -0,0 +1,153 @@
...
+     * A good place to keep a *related template resource* path.
+     */
+    fluid.defaults("gpii.pcp.exemplar", {
+        gradeNames: "fluid.component",
+        mergePolicy: {
+            widgetOptions: "noexpand"
+        },
+        template: null,
+        grade: null,
+        schemeType: null,
```
amb26: Perhaps this should read "schemaType" ?

src/pcp/js/exemplars.js
```diff
@@ -0,0 +1,153 @@
...
+            },
+            stepper: {
+                type: "gpii.pcp.exemplar.stepper"
+            },
+            textfield: {
+                type: "gpii.pcp.exemplar.textfield"
+            }
+        },
+        invokers: {
+            getExemplarBySchemeType: {
```
amb26: Schema again?

src/pcp/js/index.js
```diff
@@ -0,0 +1,424 @@
...
+
+    /**
+     * Updates the passed DOM element to contain the name of the active preference
+     * set. If there is no currently active preference set (e.g. if there is no
+     * keyed-in user), nothing should be displayed.
+     * @param activeSetElement {Object} A jQuery object representing the DOM element
+     * whose text is to be updated.
+     * @param preferences {Object} An object containing all preference set, as well
+     * as information about the currently active preference set.
+     */
```
amb26: @return here

tests/IntegrationTestDefs.js
```diff
@@ -80,7 +99,17 @@ gpii.tests.app.testDefs = {
     sequence: [{ // Test the menu that will be rendered
         event: "{that gpii.app.menu}.events.onCreate",
         listener: "gpii.tests.app.testInitialMenu"
-    }, { // Test menu after key in
+    }, [ // PCP window tests
```
amb26: This is a good start at extending the existing tests - but we will need substantial new test infrastructure to verify at least the basics of the rendered UI - e.g. that it contains some kind of populated HTML element for each setting, that they can be operated to initiate a settings change message, etc.

src/pcp/js/settingsPanel.js
```diff
@@ -0,0 +1,472 @@
...
+        gpii = fluid.registerNamespace("gpii");
+
+    fluid.registerNamespace("gpii.pcp");
+
+
+    /**
+     * Creates the binding with the already rendered DOM elements.
+     * Expects: widget configuration and model
+     */
+    fluid.defaults("gpii.pcp.settingPresenter", {
```
amb26: Now this rendering system has stabilised somewhat, it should be given some in-browser unit tests - consult https://github.com/GPII/universal/tree/master/gpii/node_modules/lifecycleManager/test for our basic pattern for these

src/app.js
danailbd: I totally agree with this, but we were influenced by the framework's event names, such as `onCreate`, `onDestroy` events.
Anyway, I'll change it to past tense

georgitodorov:
@amb26, we updated this pull request to include all of the tasks we have been working on for the past few weeks (with the exception of UI tests which are yet to come): [GPII-2528](https://issues.gpii.net/browse/GPII-2528), [GPII-2634](https://issues.gpii.net/browse/GPII-2634), [GPII-2646](https://issues.gpii.net/browse/GPII-2646), [GPII-2657](https://issues.gpii.net/browse/GPII-2657), [GPII-2659](https://issues.gpii.net/browse/GPII-2659). We also fixed and created additional tests for the electron-related functionality (as part of [GPII-2636](https://issues.gpii.net/browse/GPII-2636)). 

At the moment there are no failing tests. However, the build is failing and this is due to a leak related to the waiting dialog created by @michelled. I saw that changes had been made so that the timers are cleared whenever the application is destroyed but this does not seem to be sufficient for some reason when performing tests. Could you or Michelle please take a look at this?

We would greatly appreciate it if you can review the code and let us know if there are any areas in which we can improve it. Thank you!

src/app.js
```diff
@@ -204,14 +346,306 @@ gpii.app.performQuit = function () {
...
+ * @param key {String} The name of the setting. Must be unique as
+ * subsequent requests to the GPII API will use this key as identifier.
+ * @param settingDescriptor {Object} A descriptor for the given setting
+ * containing its title, description and constraints regarding its value.
+ * @return {Object} The view model for the setting.
+ */
+gpii.app.createSettingModel = function (key, settingDescriptor) {
+    var schema = settingDescriptor.schema;
+
+    return {
```
amb26: A bit more compact will be return fluid.extend(true, {path: key, etc.}, schema);
(At this stage fluid.extend is just an alias for $.extend)
May as well not rename "enum" here for clarity - indeed, you might increase transparency by just assigning "schema" wholesale as a member of the model. In general each time we transform and rename members of payloads we lose the ability to comprehend the overall design.

src/app.js
```diff
@@ -702,15 +1464,19 @@ gpii.app.menu.generateMenuTemplate = function (/* all the items in the menu */)
     var menuTemplate = [];
     fluid.each(arguments, function (item) {
         if (item) {
-            menuTemplate.push(item);
+            if (Array.isArray(item)) {
```
amb26: Looks like this could just be a call to fluid.flatten(fluid.makeArray(arguments))

src/pcp/js/exemplars.js
```diff
@@ -0,0 +1,201 @@
...
+"use strict";
+(function (fluid) {
+    var gpii = fluid.registerNamespace("gpii");
+
+    fluid.registerNamespace("gpii.pcp");
+
+
+    /**
+     * Represents an "exemplar" (configuration) for a component.
+     * A good place to keep a *related template resource* path.
```
amb26: Expand description a bit to list the other things held by an exemplar - schema type etc. and document the resulting options structure better.

src/pcp/js/index.js
```diff
@@ -0,0 +1,490 @@
...
+    gpii.pcp.onPreferenceSetPickerDestroy = function (container) {
+        container.find("option").remove();
+    };
+
+    /**
+     * Updates the passed DOM element to contain the name of the active preference
+     * set. If there is no currently active preference set (e.g. if there is no
+     * keyed-in user), nothing should be displayed.
+     * @param activeSetElement {Object} A jQuery object representing the DOM element
```
amb26: This argument as {jQuery}

src/pcp/js/index.js
```diff
@@ -0,0 +1,490 @@
...
+            "onSettingAltered.notifyIPC": {
+                funcName: "gpii.pcp.notifySettingUpdate",
+                args: ["{arguments}.0", "{arguments}.1"]
+            }
+        },
+        invokers: {
+            "updatePreferences": {
+                changePath: "preferences",
+                value: "{arguments}.0",
+                source: "outer"
```
amb26: Very nice use of source tracking here, but do give it a more descriptive name - remember that these names might propagate around an entire design of components linked by relays. Perhaps something like pcp.mainWindow

src/pcp/js/index.js
```diff
@@ -0,0 +1,490 @@
+/*
```
amb26: Factor this file a bit better. For example it looks like it could contain "header", "splash", "clientChannel" or so. 
Also it would be helpful if the files in this module were clearly separated according to whether they were attached to a renderer or main process.
The comment contains a few stray CRs, looks like it entered a faulty file conversion/paste from DOS format

src/app.js
```diff
@@ -277,6 +721,117 @@ fluid.onUncaughtException.addListener(function (err) {
...
+fluid.defaults("gpii.app.pcp", {
+    gradeNames: "fluid.modelComponent",
+
+    model:  {
+        keyedInUserToken: null
+    },
+
+    /*
+     * Options to be passed to the Electron `BrowserWindow`.
+     * Only not reserved options can be altered. See `reservedAttrs`.
```
amb26: Add a link to the relevant docs here to help the readear

src/app.js
```diff
@@ -277,6 +721,117 @@ fluid.onUncaughtException.addListener(function (err) {
...
+        frame: false,
+        fullscreenable: false,
+        resizable: false,
+        alwaysOnTop: true,
+        skipTaskbar: true,
+        backgroundColor: "transparent"
+    },
+
+    members: {
+        pcpWindow: {
```
amb26: This is simple enough to be usefully expressed in the compact format - 
pcpWindow: "@expand:gpii.app.pcp.makePCPWindow({that}.options.attrs)"

src/app.js
```diff
@@ -663,37 +1342,120 @@ gpii.app.menu.getUserName = function (userToken) {
     return name;
 };
 
+
+/**
+*  Object representing options for a `Electron` `ContextMenu` item.
+ * @typedef {Object} ElectronMenuItem
```
amb26: Very nice - now all we need is ... TYPESCRIPT... or some similar abomination : P

src/app.js
```diff
@@ -12,19 +12,24 @@ https://github.com/GPII/universal/blob/master/LICENSE.txt
 */
 "use strict";
 
+var electron = require("electron");
```
amb26: I think the time has finally come that this single-file app can be more helpfully factored. It looks like there are at least files "menu.js", "tray.js" etc. and see what else you can extract. Also, could you create subdirectories in "src" to separately group together those files which are used in "main" and "renderer" processes. (and a "shared" if there happen to be any definitions shared between the two processes)

package.json
```diff
@@ -23,10 +23,11 @@
...
     "electron": "1.4.1",
-    "request": "2.69.0",
+    "electron-edge": "6.5.4",
+    "gpii-windows": "git://github.com/danailbd/windows.git#hst-2017",
     "infusion": "3.0.0-dev.20170830T182157Z.392b2f8",
-    "gpii-windows": "git://github.com/GPII/windows.git#hst-2017",
     "node-jqunit": "1.1.7",
-    "electron-edge": "6.5.4"
+    "request": "2.69.0",
+    "winstrap": "^0.5.11"
```
amb26: Our standard is to always bind to exact versions of dependencies to avoid injection of security flaws, etc.

src/pcp/js/widgets.js
```diff
@@ -0,0 +1,304 @@
...
+        components: {
+            slider: {
+                options: {
+                    listeners: {
+                        "onCreate.bindSlideEvt": {
+                            "this": "{that}.container",
+                            "method": "on",
+                            "args": ["input", "{that}.onSlide"]
+                        },
+                        "onCreate.bindRangeChangeEvt": {
```
amb26: I'm not sure the namespace for this event is descriptive - perhaps it is left over from a previous definition?

src/pcp/js/widgets.js
```diff
@@ -0,0 +1,304 @@
...
+
+/* global fluid */
+
+"use strict";
+(function (fluid) {
+    var gpii = fluid.registerNamespace("gpii");
+
+    fluid.registerNamespace("gpii.pcp.widgets");
+
+    gpii.pcp.widgets.noop = function () {
```
amb26: We can axe this - if you need a general function that does nothing you can use fluid.identity - but it looks like the only use is as a click binder in an invoker, in which case you can use fluid.notImplemented since I imagine you want to force the user to override this

src/pcp/js/widgets.js
georgitodorov: Actually, we are replacing the `onCreate.bindRangeChangeEvt` listener which comes from the default configuration of the `fluid.textfieldSlider` component. We do this in order not to constantly update the slider's model while the user is dragging the thumb. The update occurs when the user lets go of the thumb.

src/main/menu.js
```diff
@@ -0,0 +1,420 @@
...
+        }
+    },
+    menuLabels: {
+        psp: "Open PSP",
+        keyedIn: "Keyed in with %snapsetName",    // string template
+        keyOut: "Key-out of GPII",
+        notKeyedIn: "(No one keyed in)",
+
+        // XXX DEV items
+        exit: "Exit GPII",
```
amb26: We should factor these dev items into a separate grade so it is easier to produce production builds of the app

src/renderer/psp/js/header.js
```diff
@@ -0,0 +1,174 @@
...
+https://github.com/GPII/universal/blob/master/LICENSE.txt
+*/
+
+/* global fluid */
+
+"use strict";
+(function (fluid) {
+    var gpii = fluid.registerNamespace("gpii");
+
+    fluid.defaults("gpii.psp.header", {
```
amb26: More documentation on this file and grade would be helpful

src/renderer/psp/js/index.js
```diff
@@ -0,0 +1,58 @@
...
+                        onSettingAltered: "{clientChannel}.alterSetting",
+                        onActivePreferenceSetAltered: "{clientChannel}.alterActivePreferenceSet",
+                        onContentHeightChanged: "{clientChannel}.changeContentHeight"
+                    }
+                }
+            }
+        }
+    });
+
+    $(function () {
```
amb26: It may seem perverse, but I'd like to see this function invocation moved into a 1-line <script tag at the base of the markup instead. It helps reuse a lot if one can separate the process of grade definition and instantiation into different files.

src/renderer/psp/js/widgets.js
```diff
@@ -0,0 +1,291 @@
...
+
+    fluid.defaults("gpii.psp.widgets.dropdown", {
+        gradeNames: ["gpii.psp.widgets.attrsExpander", "fluid.rendererComponent"],
+        model: {
+            optionNames: [],
+            optionList: [],
+            selection: null
+        },
+        modelListeners: {
+            "*": {
```
amb26: To avoid unnecessary notifications, this path should be "" rather than "*" - https://docs.fluidproject.org/infusion/development/ChangeApplierAPI.html#wildcards-in-model-path-references

tests/IntegrationTests.js
```diff
@@ -84,4 +83,13 @@ gpii.tests.app.bootstrapServer = function (testDefs, transformer) {
     return kettle.test.bootstrap(testDefs, fluid.makeArray(transformer).concat([gpii.tests.app.testDefToServerEnvironment]));
 };
 
-gpii.tests.app.bootstrapServer([fluid.copy(gpii.tests.app.testDefs), fluid.copy(gpii.tests.dev.testDefs)]);
+gpii.tests.app.bootstrapServer([
+    /*
```
amb26: Looks like this comment can be removed now?

src/renderer/psp/js/index.js
georgitodorov: I moved the definition of the `gpii.psp` component to a separate file and left its instantiation here. This is equivalent to putting a script tag at the end of the body tag in the markup. I always refrain from placing JavaScript code directly in a script tag within an HTML file unless it is absolutely necessary. In this particular case, I do not think that such a necessity is present.

amb26:
ok to test

src/renderer/psp/js/heightChangeListener.js
```diff
@@ -0,0 +1,33 @@
...
+        gradeNames: ["fluid.viewComponent"],
+        listeners: {
+            "onCreate.initResizeListener": {
+                "this": "@expand:$({that}.container.0.contentWindow)",
+                method: "on",
+                args: ["resize", "{that}.onHeightChanged"]
+            }
+        },
+        invokers: {
+            onHeightChanged: fluid.notImplemented
```
amb26: An event would be a more idiomatic approach than a non-implemented invoker

src/renderer/psp/js/header.js
```diff
@@ -0,0 +1,182 @@
...
+                args: ["{change}.value"],
+                excludeSource: ["init", "psp.mainWindow"]
+            },{
+                funcName: "gpii.psp.updateActiveSetElement",
+                args: ["{that}.dom.activePreferenceSet", "{that}.model.preferences"]
+            }]
+        },
+        components: {
+            preferenceSetPicker: {
+                type: "@expand:gpii.psp.getPreferenceSetPickerType({that}.model.preferences.sets)",
```
amb26: Pretty cool parameterised dynamic subcomponent type :)

src/main/utils.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+};
+
+/**
+ * Checks whether two objects are equal by using `JSON.stringify` internally.
+ * This  means that if the compared values are arrays or objects, the ordering
+ * of the elements/properties within them matters.
+ * @return {Boolean} `true` if the values are equal and `false` otherwise.
+ */
+app.equalsAsJSON = function (obj1, obj2) {
```
amb26: Use of this utility is unstable since even the latest language standard doesn't specify that the property iteration order will be stable in all cases: https://stackoverflow.com/questions/30076219/does-es6-introduce-a-well-defined-order-of-enumeration-for-object-properties - please rewrite calls to this using fluid.model.diff instead - https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/DataBinding.js#L1196

src/main/utils.js
```diff
@@ -0,0 +1,92 @@
...
+You may obtain a copy of the License at
+https://github.com/GPII/universal/blob/master/LICENSE.txt
+*/
+"use strict";
+
+var os       = require("os");
+var fluid    = require("infusion");
+var electron = require("electron");
+
+var app = fluid.registerNamespace("gpii.app");
```
amb26: Please rewrite this so that the namespace usage is consistent with other files - even though this creates an annoying redundancy:
    var gpii = fluid.registerNamespace("gpii");
    fluid.registerNamespace("gpii.app");

    gpii.app.isWin10OS = function () { ... etc

src/main/utils.js
```diff
@@ -0,0 +1,92 @@
+/*!
+GPII Application
```
amb26: This comment is now outdated

src/main/tray.js
```diff
@@ -0,0 +1,163 @@
+/*!
+GPII Application
```
amb26: Please update all these file comments to reflect file contents plus a short description of file's purpose

src/main/restartDialog.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * A component that serves as simple interface for communication with the
+ * electron `BrowserWindow` restart dialog.
+ */
+fluid.defaults("gpii.app.dialog.restartDialog.channel", {
+    gradeNames: ["fluid.component"],
+
+    events: {
+        // onClosed: null
```
amb26: If this grade is intended for use as a mixin grade for another, specify what that base grade is (holding definition of onClosed). In these cases we also tend to lead out the "gradeNames" specification (if the grade cannot function as a self-sufficient component)

src/main/tray.js
```diff
@@ -0,0 +1,163 @@
...
+    members: {
+        tray: {
+            expander: {
+                funcName: "gpii.app.makeTray",
+                args: ["{that}.options.icons.keyedOut", "{psp}.show"]
+            }
+        }
+    },
+    icons: {
+        pendingChanges: "../icons/gpii-pending.png",
```
amb26: Could you update these references to support and use package-relative paths such as "%gpii-app/icons/gpii-pending.png" - this will make the work easier in future of migrating such assets into their own module as has been requested. You can achieve this by passing such paths to https://docs.fluidproject.org/infusion/development/NodeAPI.html#fluidmoduleresolvepathpath in order to get a full filesystem path

src/main/tray.js
```diff
@@ -0,0 +1,163 @@
...
+};
+
+/**
+ * Returns the path to the icon for the Electron Tray based on whether there is a
+ * keyed-in user and on the pending setting changes (if any).
+ * @param keyedInUserToken {String} The token if the keyed-in user or `null` if
+ * there is no such.
+ * @param pendingChanges {Array} An array containing all pending setting changes.
+ * @param icons {Object} An object containing all possible icon paths.
+ * @return The tooltip label for the Electron Tray.
```
amb26: Looks like this return description is incorrect

src/main/tray.js
```diff
@@ -0,0 +1,163 @@
...
+  * whenever the user left clicks on the tray icon or uses the PSP window shortcut.
+  */
+gpii.app.makeTray = function (icon, openPSP) {
+    var tray = new Tray(path.join(__dirname, icon));
+
+    tray.on("click", function () {
+        openPSP();
+    });
+
+    globalShortcut.register("Super+CmdOrCtrl+Alt+U", function () {
```
amb26: Please move this shortcut string into component configuration

tests/RestartDialogTestDefs.js
```diff
@@ -0,0 +1,285 @@
...
+                "{that}.app.settingsBroker.model.pendingChanges"
+            ]
+        }, {
+            funcName: "jqUnit.assertFalse",
+            args: [
+                "RestartDialog: Dialog is hidden on RestartNow in PSP",
+                "@expand:{that}.app.restartDialog.isShown()"
+            ]
+        }
+        // TODO check whether setting applied event was fired
```
amb26: Can you address this TODO here? Or does it require migration to @kaspermarkus 's branch (which perhaps we are already using)

tests/SettingsBrokerTestDefs.js
```diff
@@ -0,0 +1,231 @@
...
+require("../src/main/app.js");
+
+fluid.registerNamespace("gpii.tests.settingsBroker.testDefs");
+
+var liveSettingChange = {
+    path: "textfieldPath",
+    oldValue: "Old value",
+    value: "New Value",
+
+    icon: "../../../../icons/gear-cloud-white.png",
```
amb26: As with the icon defs above, can we move these towards the package-relative form e.g. "%gpii-app/icons/gear-cloud-white.png"

src/renderer/psp/js/psp.js
```diff
@@ -0,0 +1,62 @@
...
+Seventh Framework Programme (FP7/2007-2013) under grant agreement no. 289016.
+You may obtain a copy of the License at
+https://github.com/GPII/universal/blob/master/LICENSE.txt
+*/
+
+/* global fluid */
+
+"use strict";
+(function (fluid) {
+    var gpii = fluid.registerNamespace("gpii");
```
amb26: This triggers a linting failure as unused - the entire line can be removed

tests/SettingsBrokerTestDefs.js
georgitodorov: The exact value of the `icon` property (and whether it points to a file which exists or not) does not matter here. For the purpose of the tests it is important that each setting has an `icon` property and that the `src` of its image is the same as the value of the property. As the `fluid.module.resolvePath` does not work properly in this version of the UI tests, I am thinking of not modifying these values for now.

src/main/gpiiConnector.js
```diff
@@ -0,0 +1,280 @@
...
+
+
+/**
+ * A decorator for the extracted preferences that applies values that are to be used
+ * for development.
+ */
+gpii.app.dev.gpiiConnector.mockPreferences = function (preferences) {
+    function applyManualLivenessFlag(settings) {
+        settings.forEach(function (setting) {
+            // XXX a workaround as the Magnifier settings are missing the `solutionName` property
```
amb26: What will I do when I am no longer able to put such questions to @kaspermarkus ....

tests/SettingsBrokerTestDefs.js
amb26: Can you say why `fluid.module.resolvePath` does not work properly?

src/main/dialog.js
```diff
@@ -0,0 +1,163 @@
...
+            suffixPath: suffixPath
+        });
+};
+
+/**
+ * Creates a dialog. This is done up front to avoid the delay from creating a new
+ * dialog every time a new message should be displayed.
+ * @param windowOptions {Object} The raw Electron `BrowserWindow` settings
+ * @param position {Object} The desired position for the component
+ * @param positoins.x {Number}
```
amb26: Typo here

src/main/menu.js
```diff
@@ -0,0 +1,452 @@
...
+        click: "onPSP",
+        args: {
+            token: keyedInUserToken
+        }
+    };
+};
+
+/**
+  * Creates a JSON representation of a menu.
+  * @param {Object} An object containing a menu item template.
```
amb26: Looks like this comment is not quite right, as the function seems to accept a set of "splatted" arguments

src/main/menu.js
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    return preferenceSetsLabels;
+};
+
+/**
+  * Generates an object that represents the menu items for opening the settings panel
+  * @param keyedInUserToken {String} The user token that is currently keyed in.
+  * @param openSettingsStr {String} The string to be displayed for the open setting panel menu item.
+  * @returns {ElectronMenuItem}
```
amb26: @return

src/main/restartDialog.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+fluid.defaults("gpii.app.dialog.restartDialog", {
+    gradeNames: ["gpii.app.dialog"],
+
+    invokers: {
+        showIfNeeded: {
+            funcName: "gpii.app.dialog.restartDialog.showIfNeeded",
+            args: ["{that}", "{arguments}.0"]
+        },
+        isShown: {
```
amb26: Rather than reaching into the impl here, make this a standard modelised element of the base class. That is, place "isShown" into a model field, i) isShown invoker is removed in favour of the client examining this field, ii) show and hide are just model listeners rather than invokers

src/main/dialog.js
```diff
@@ -0,0 +1,163 @@
...
+                "{that}.options.config.attrs.width",
+                "{that}.options.config.attrs.height"
+            ]
+        },
+        resetWindowPosition: {
+            funcName: "gpii.app.setWindowPosition",
+            args: ["{that}.dialog", "@expand:{that}.getWindowPosition()"]
+        },
+        // Simple default behaviour
+        show: {
```
amb26: Turn these into namespaced modelListener to the "isShown" field as described below in restartDialog comments

src/main/dialog.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+    },
+    listeners: {
+        "onDestroy.cleanupElectron": {
+            this: "{that}.dialog",
+            method: "destroy"
+        }
+    },
+    invokers: {
+        getWindowPosition: {
```
amb26: Rename to "getInitialWindowDimensions"

src/main/utils.js
```diff
@@ -0,0 +1,84 @@
...
+ */
+gpii.app.isWin10OS = function () {
+    var osRelease = os.release(),
+        delimiter = osRelease.indexOf("."),
+        majorVersion = osRelease.slice(0, delimiter);
+    return majorVersion === "10";
+};
+
+/**
+* Get the position of `Electron` `BrowserWindows`
```
amb26: This could do with some further explanation/renaming - it looks like it gets the *desired* window position given its dimensions

configs/app.testing.json
```diff
@@ -0,0 +1,19 @@
...
+                "target": "{that flowManager}.options.gradeNames"
+            },
+            "distributePcpChannelConnector": {
+                "record": "gpii.app.dev.gpiiConnector",
+                "target": "{that gpiiConnector}.options.gradeNames"
+            }
+        }
+    },
+    "mergeConfigs": [
+        "../node_modules/universal/gpii/configs/gpii.config.all.development.dr.production.json"
```
amb26: Do use the %universal syntax here

src/main/restartDialog.js
georgitodorov: Our assumption for having a show/hide invokers was that the logic for showing/hiding of a dialog could be more complex. Take for instance the PSP itself (even though it is not a dialog). When showing it, we actually move it to the screen, focus it and change a property in the model. And what if there is a need to pass a parameter when showing a dialog (like when displaying a survey)? Should we have an invoker in that case that sets the parameter to the component’s options or model, then changes the `isShown` model property and leaves it up to the model listener to handle the showing operation taking into account the set parameter?

Maybe we are accustomed to using functions/methods for such operations in other frameworks/programming languages instead of modifying the intrinsic object properties. Is this not applicable in this particular case for Infusion components?

yuriy-donev: @amb26, I just wanted to check if you have had the chance to review Georgi's comment. It's the largest remaining comment from the original PR and I wanted to get some sense on the remaining work depending on what needs to be done here. Thank you!

tests/SettingsBrokerTestDefs.js
georgitodorov: Actually, it seems that in the PSP integration tests `fluid.module.resolvePath` works fine. I wanted to use this function in the PSP UI tests, as well, and there it was not working properly so I didn't change the paths that appeared there.

src/main/waitDialog.js
```diff
@@ -0,0 +1,124 @@
...
+        },
+        fileSuffixPath: "waitDialog/index.html"
+    },
+    model: {
+        dialogMinDisplayTime: 2000, // minimum time to display dialog to user in ms
+        dialogStartTime: 0, // timestamp recording when the dialog was displayed to know when we can dismiss it again
+        timeout: 0
+    },
+    modelListeners: {
+        isShown: {
```
amb26: Yup, I think this is now great

PR #32, opened 2017-11-17
GPII-2294: In-process user listeners

stegru:


stegru:
ok to test

stegru:
ok to test

stegru:
ok to test

stegru:
ok to test (making sure changes to universal didn't break anything)

stegru:
ok to test

stegru:
okk to test

stegru:
ok to test

stegru:
ok to test

amb26:
It's the blasted "dialog width" test again - @georgitodorov , have you run into this in your own branches and have a fix? I seem to remember that in your current pull you have disabled the dialog width test? I wonder what the cause is

stegru:
I wonder if this is caused by the display settings handler tests setting the resolution to 800 and/or the DPI to the maximum.

These two values aren't set at the same time, there could be some weird state where the screen width is still 800px (because the resolution hasn't finished adjusting), and the DPI is set to 200%, it would explain the 400px wide window.

This assumes the window is created before the tests, and electron/Windows keeps an eye on the available desktop size and shrinks the window to fit.

stegru:
ok to test (last one timed out)

stegru:
ok to test

PR #34, opened 2018-01-29
Gpii 2732

danailbd:
This PR introduces the survey pop-up functionality which includes:
* Static survey payloads (currently loaded from config files)
* Implementation of a survey pop-up
    * electron BrowserWindow
    * embedding of Qualtrics survey inside a webview
* Simple time-based trigger manager that waits for timers to finish to indicate success on a condition

For further details please refer to the related tickets' comments,  the related [Technical Evaluation Document](https://wiki.gpii.net/w/Technology_Evaluation_-_Survey_Functionality) and components documentation.

Related tickets:
* [GPII-2755](https://issues.gpii.net/browse/GPII-2755)
* [GPII-2758](https://issues.gpii.net/browse/GPII-2758)
* [GPII-2759](https://issues.gpii.net/browse/GPII-2759)
* [GPII-2760](https://issues.gpii.net/browse/GPII-2760)
* [GPII-2762](https://issues.gpii.net/browse/GPII-2762)
* [GPII-2763](https://issues.gpii.net/browse/GPII-2763)
* [GPII-2764](https://issues.gpii.net/browse/GPII-2764)
* [GPII-2776](https://issues.gpii.net/browse/GPII-2776)


danailbd:
Reopened from the [inner PR](https://github.com/danailbd/gpii-app/pull/22). Link it in order to keep the comments history.

src/main/app.js
```diff
@@ -46,8 +50,13 @@ require("electron").app.on("ready", gpii.app.electronAppListener);
 // Override default behaviour - don't exit process once all windows are closed
 require("electron").app.on("window-all-closed", fluid.identity);
 
-/*
- ** Component to manage the app.
+/**
+ * A component to manage the app. Each subcomponent of the app (except for the first
```
amb26: This shouldn't be necessary, and depends on a framework facility that will be withdrawn shortly. If a component makes use of another component's facilities, (e.g. factsManager -> installID) this should be expressed either through an explicit dependency, or else mediated via an event or options distribution

src/main/app.js
```diff
@@ -59,37 +68,96 @@ fluid.defaults("gpii.app", {
...
+            type: "gpii.installID",
+            priority: "after:networkCheck"
+        },
+        factsManager: {
+            type: "gpii.app.factsManager",
+            priority: "after:installID"
+        },
+        rulesEngine: {
+            type: "gpii.app.rulesEngine",
+            priority: "after:factsManager",
```
amb26: For example, this annotation is unnecessary since the options for the component contain an explicit reference to the factsManager which guarantees that it will be available

src/main/factsManager.js
```diff
@@ -0,0 +1,235 @@
...
+};
+
+/**
+ * Get facts collected from all fact providers.
+ * @param that {Component} The gpii.app.factsManager component
+ * @return {Object} A map of all facts - { factName: factValue, ... },
+ * where the `factName` is defined in the corresponding provider
+ */
+gpii.app.factsManager.getFacts = function (that) {
+    var factProviders = gpii.app.getSubcomponents(that);
```
amb26: To avoid races of the kind you are worried about, it would be clearer if this relationship was mediated by an options distribution from the factManager to the onCreate event of the fact providers, which directs them to register themselves with the factManager. The cleanest expression would just write this in a grade which is distributed onto the factProviders.

src/main/app.js
```diff
@@ -59,37 +68,96 @@ fluid.defaults("gpii.app", {
...
     },
-    // prerequisites
+    members: {
+        machineId: "@expand:{that}.installID.getMachineID()"
+    },
     components: {
-        psp: {
-            type: "gpii.app.psp",
-            createOnEvent: "onPrerequisitesReady",
+        networkCheck: { // Network check component to meet GPII-2349
```
amb26: This component can be removed since it was only required for the HST testing deployment. It got into the main branch through a faulty merging of build branches.

src/main/app.js
```diff
@@ -59,37 +67,83 @@ fluid.defaults("gpii.app", {
...
+            priority: "after:installID"
+        },
+        surveyManager: {
+            type: "gpii.app.surveyManager",
+            priority: "after:factsManager"
+        },
+        dialogManager: {
+            type: "gpii.app.dialogManager",
+            createOnEvent: "onPSPPrerequisitesReady",
+            priority: "after:surveyManager",
```
amb26: This annotation can be removed since the component's definition includes a dependency on the surveyManager - in addition, this component is createOnEvent and so will definitely be constructed later

src/main/app.js
```diff
@@ -59,37 +67,83 @@ fluid.defaults("gpii.app", {
...
+        installID: {
+            type: "gpii.installID"
+        },
+        factsManager: {
+            type: "gpii.app.factsManager",
+            priority: "after:installID"
+        },
+        surveyManager: {
+            type: "gpii.app.surveyManager",
+            priority: "after:factsManager"
```
amb26: These two priorities should be removed as well - I don't see a hard dependency between the surveyManager and the factsManager - what would happen if the latter constructed first?

src/main/surveys/surveyConnector.js
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Should be called when a trigger's conditions are met. As a result, a static
+ * payload (with keyedInUserToken and machineId added as query paramenters) for
+ * the survey to be displayed will be sent via the `onSurveyRequired` event.
+ * @param that {Component} The `gpii.app.staticSurveyConnector` instance.
+ */
+gpii.app.staticSurveyConnector.notifyTriggerOccurred = function (that) {
+    var fixture = that.options.config.surveyFixture;
+    fixture.url = fluid.stringTemplate(fixture.url +
+        "?keyedInUserToken=%keyedInUserToken&machineId=%machineId", that.model);
```
amb26: Place this url template into the component's options

src/renderer/survey/js/survey.js
```diff
@@ -0,0 +1,49 @@
...
+/* global fluid */
+
+"use strict";
+/**
+ * A wrapper component for the `gpii.survey.channel` (which handles the
+ * communication between the `BrowserWindow` and the main electron process)
+ * and the `gpii.survey.popup` (which wraps and serves as an intermediary
+ * for the webview).
+ */
+(function (fluid) {
```
amb26: File closure seems unnecessary

src/renderer/survey/js/webview.js
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Checks whether a given DOM element is a 'break out' link, i.e.
+     * whether it is an anchor tag whose target is blank and which has
+     * the `flc-breakOut` class. The `flc-breakOut` class helps the
+     * application distinguish between regular links that should open in a
+     * new browser window and the 'break out' link which should also
+     * close the survey pop-up.
+     * @param target {Element} The DOM element to be checked.
+     */
+    function isBreakOutLink(target) {
```
amb26: Please add some bare-bones namespacing to this file so that this library of functions can expand gracefully - e.g. something like var gpii = gpii || {webview: {}}
And then the contents of this file can be registered as, e.g.
gpii.webview.isBreakoutLink = function (target) {

etc.

src/renderer/survey/js/webview.js
```diff
@@ -0,0 +1,132 @@
+/**
+ * The preload script for the survey webview
```
amb26: Can you think of any way that the contents of this file could be tested? Perhaps via some kind of bare-bones "mock survey server"?

src/renderer/survey/js/survey.js
georgitodorov: As agreed on a call with Antranig on 13 Feb, this comment will not be addressed now. The reasons for this are that there will be huge diffs (because indentation will need to be corrected) as closure is used everywhere in the `renderer` part of the project. Also, there are a few branches which have stemmed either from this one or from master and there will be conflicts when merging them due to the indentation changes.

src/renderer/survey/js/webview.js
georgitodorov: There is no need for a mock HTTP server in order to test the webview preload script. The basic idea of the approach which we have used can be found in the following JIRA ticket: https://issues.gpii.net/browse/GPII-2870.

package.json
```diff
@@ -24,7 +24,7 @@
   "dependencies": {
     "electron": "1.4.1",
     "electron-edge": "6.5.4",
-    "gpii-windows": "git://github.com/danailbd/windows.git#hst-2017",
+    "gpii-windows": "git://github.com/danailbd/windows.git",
```
amb26: Is there something special in this branch which is not yet in trunk? We should try to get merged up

danailbd: The problem is that the master branch of GPII-windows repo is [pointing to some commit](https://github.com/GPII/windows/blob/master/package.json#L15) of `universal` that is missing the `pcpChannel`, instead of master. In our fork, we've change it to use master. Probably we'd want to apply this change in the main windows repo as well?

amb26: This got updated recently, can we recheck it?

package.json
```diff
@@ -24,7 +24,7 @@
...
     "infusion": "3.0.0-dev.20170830T182157Z.392b2f8",
     "node-jqunit": "1.1.7",
```
amb26: Please re-merge up with trunk of gpii-app to take in recent dependencies plus renamed repository gpii-universal

amb26:
ok to test

amb26:
The tests pass - but there is still actually a failure in the logs. Starting the settings broker integration tests (at line 206541 of my logs)

    07:18:04.234:  Successfully queued test Settings broker integration tests

eventually leads to (at line 211882 of my logs):

    07:18:12.519:  FATAL ERROR: Uncaught exception: Cannot read property 'execute' of undefined
TypeError: Cannot read property 'execute' of undefined
    at that.execute (\\vboxsrv\vagrant\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:599:25)

I believe this will be due to still having two passive fixtures in a row in SettingsBrokerTestDefs.js 

We should investigate separately why this doesn't lead to a test failure (also, this may be the reason that I find the magnifier still enabled when the tests finish), but in the meantime could you rewrite these tests so that they don't cause the logged failure - thanks!

PR #35, opened 2018-02-09
GPII-2568: Updated references to universal with gpii-universal

javihernandez:
Works with https://github.com/GPII/windows/pull/161

### DIFF
--- a/configs/app.dev.json
+++ b/configs/app.dev.json
@@ -18,6 +18,6 @@
         }
     },
     "mergeConfigs": [
-        "%universal/gpii/configs/gpii.config.development.all.local.json"
+        "%gpii-universal/gpii/configs/gpii.config.development.all.local.json"
      ]
 }
--- a/configs/app.production.json
+++ b/configs/app.production.json
@@ -10,6 +10,6 @@
         }
     },
     "mergeConfigs": [
-        "%universal/gpii/configs/gpii.config.all.development.dr.production.json"
+        "%gpii-universal/gpii/configs/gpii.config.all.development.dr.production.json"
      ]
 }
--- a/configs/app.testing.json
+++ b/configs/app.testing.json
@@ -14,6 +14,6 @@
         }
     },
     "mergeConfigs": [
-        "%universal/gpii/configs/gpii.config.all.development.dr.production.json"
+        "%gpii-universal/gpii/configs/gpii.config.all.development.dr.production.json"
      ]
 }
--- a/main.js
+++ b/main.js
@@ -16,7 +16,7 @@ var fluid = require("infusion"),
     gpii = fluid.registerNamespace("gpii"),
     kettle = fluid.registerNamespace("kettle");
 
-require("universal");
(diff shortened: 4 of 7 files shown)


javihernandez:
> CI job failed: https://ci.gpii.net/job/gpii-app-tests/166/

@stegru, @amb26, not sure why, but I'm getting this error, which comes from the metrics code and during the execution of the UserLogonStateChange tests.

`08:07:52.847:  FATAL ERROR: Uncaught exception: Cannot read property 'model' of undefined

TypeError: Cannot read property 'model' of undefined at gpii.metrics.logonStateChanged (\\vboxsrv\vagrant\node_modules\gpii-universal\gpii\node_modules\eventLog\src\metrics.js:112:51)`

Any ideas? Meanwhile, I'll keep looking into it.

stegru:
Without fully understanding the reason behind it, here's a fix: https://github.com/stegru/universal/commit/848b0011a245adcec1c38e97e5680b13e99c439f

`lifecycleManager.getSession()` returns null, perhaps the token passed to it (the oldValue of logonChange) was some test data.

javihernandez:
@stegru, thanks for taking a look at this.

Right, _getSession_ returns _null_ and I can confirm that your fix works fine.
However, I've also checked the values of _newValue_ and _oldValue _ and I just wanted to share them with you.

`
"newValue": {
    "inProgress": false,
    "timeStamp": 1518529806748,
    "type": "logout",
    "userToken": undefined
}

"oldValue": {
    "inProgress": true,
    "timeStamp": 1518529806527,
    "type": "login",
    "userToken": "bogusToken"
}
`
@amb26, I wonder if this is the expected behavior, to perform a logout of an _undefined_ user in the case where the login process can't be completed.

amb26:
No, it is not, and the test shouldn't pass even with @stegru  's fix ...

amb26:
OK - this firmly falls into the category of "this has never worked" :) It turns out that the previous trunk of gpii-app only passed its tests because it bound to a version of gpii-windows that did not have the metrics component in it. Also - there is a fault in the core definition of the gpii.metrics component at https://github.com/GPII/universal/blob/master/gpii/node_modules/eventLog/src/metrics.js#L27 since it does not itself derive from fluid.modelComponent which means that the modelListener at line 52 does not get processed - unless this grade is added via gpii.metrics.windows being applied. And it turns out the first time it was ever applied was in this branch, which references an up to date gpii-windows. @javihernandez  - can you update your pull https://github.com/GPII/universal/pull/548 so that
i) on line 27 of metrics.js we derive from fluid.modelComponent rather than fluid.component
ii) that you incorporate stegru's null checker at https://github.com/stegru/universal/commit/848b0011a245adcec1c38e97e5680b13e99c439f with a comment to explain that the reason that the lifecycleManager's session may be null is that we have logged on with a bogus token which terminated the request in error before the session could be created. And then we will feed it through the system again - thanks!

javihernandez:
ok to test

PR #39, opened 2018-02-26
GPII-2665: Create a new mechanism for rendering PSP errors

georgitodorov:
This PR includes the following changes:

- [GPII-2845](https://issues.gpii.net/browse/GPII-2845): Re-style error message dialog in PSP
- [GPII-2871](https://issues.gpii.net/browse/GPII-2871): Display PSP error message dialogs sequentially
- Unifying the styles for all dialogs in the PSP (wait, survey and error dialogs)
- A bullet list of the applications which need to be restarted in the restart dialog
- Code improvements.


javihernandez:
ok to test

amb26:
ok to test

src/main/app.js
```diff
@@ -63,6 +62,30 @@ fluid.defaults("gpii.app", {
             activeSet: null
         }
     },
+    errorsDescriptionMap: {
```
amb26: This would be clearer in its own grade

src/main/app.js
```diff
@@ -96,189 +109,156 @@ fluid.defaults("gpii.app", {
...
-         * A helper component used as mediator for handling communication
-         * between the PSP and gpiiConnector components.
-         */
-        channelMediator: {
-            type: "fluid.component",
-            createOnEvent: "onPSPPrerequisitesReady",
-            options: {
+                    keyedInUserToken: "{app}.model.keyedInUserToken"
+                },
+                modelListeners: {
```
amb26: I think this material configuring the PSP has now become too large to be declared inline in the app. I suggest that this binding material should be factored off into an "incomplete grade" named something like "pspInApp". In strict dependency terms that grade ought to stay in this file but it wouldn't be a blunder to put it into psp.js either, see what your taste is

src/main/app.js
```diff
@@ -355,9 +335,13 @@ fluid.defaults("gpii.app", {
             funcName: "gpii.app.exit",
             args: "{that}"
         },
-        "handleUncaughtException": {
+        handleUncaughtException: {
```
amb26: This could go into the dedicated "errors" component along with the descriptionMap

src/renderer/common/js/heightChangeListener.js
```diff
@@ -0,0 +1,83 @@
...
+    var gpii = fluid.registerNamespace("gpii");
+    /**
+     * Responsible for detecting height changes in the DOM element in which
+     * the component's container is nested.
+     */
+    fluid.defaults("gpii.psp.heightChangeListener", {
+        gradeNames: ["fluid.viewComponent"],
+        listeners: {
+            "onCreate.initResizeListener": {
+                "this": "@expand:$({that}.container.0.contentWindow)",
```
amb26: Binding to the $ symbol as a global is generally dangerous in a browser context since this is intended to be a lightweight symbol which can be rebound within a closure to whichever library is most convenient for the local implementation. I know that you have 100% control over the loading context within the markup but I suggest that just for general clarity this is written as "jQuery"

src/renderer/common/js/heightChangeListener.js
```diff
@@ -0,0 +1,83 @@
...
+                method: "on",
+                args: ["resize", "{that}.events.onHeightChanged.fire"]
```
amb26: We will get extra firings of this event due to any change in dimension - in order to live up to its contract, you might want to modelise the height and then bind the outward-facing event to that.

src/renderer/common/js/heightChangeListener.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    /**
+     * Computes the size of the content and fires the `onHeightChanged` event.
+     * @param that {Component} The `gpii.psp.heightObservable` instance.
+     */
+    gpii.psp.heightObservable.onContentHeightChanged = function (that) {
+        var container = that.container;
+        // get speech triangle size, in case such exists
+        var triangleHeight = container.find(".fl-speech-triangle").outerHeight(true) || 0,
```
amb26: This odd hard-binding to the class should be optionised - or the algorithm generalised in some other kind of way

src/renderer/common/js/heightChangeListener.js
```diff
@@ -0,0 +1,83 @@
...
+        gradeNames: ["fluid.viewComponent"],
+        listeners: {
+            "onCreate.initResizeListener": {
+                "this": "@expand:$({that}.container.0.contentWindow)",
+                method: "on",
+                args: ["resize", "{that}.events.onHeightChanged.fire"]
+            }
+        },
+        events: {
+            onHeightChanged: null
```
amb26: In turn, the contract for this event should be explained - i.e. when it fires and with what arguments

src/renderer/common/js/heightChangeListener.js
```diff
@@ -0,0 +1,83 @@
...
+        }
+    });
+
+    /**
+     * A `viewComponent` which continuously watches for changes in the height
```
amb26: This comment properly belongs on the other component, since that is the one which continuously watches for changes in its container. This one uses that in order to bind it to ".flc-contentHeightChangeListener" etc.

src/renderer/common/js/heightChangeListener.js
```diff
@@ -0,0 +1,83 @@
...
+     * of its container. Whenever a height modification is detected, the
+     * `onHeightChanged` event will be fired with the new height as an argument.
+     */
+    fluid.defaults("gpii.psp.heightObservable", {
+        gradeNames: "fluid.viewComponent",
+
+        selectors: {
+            heightChangeListener: ".flc-contentHeightChangeListener"
```
amb26: A name which might puzzle someone looking at the markup. Perhaps contentHeightObserved or so?

src/renderer/errorDialog/js/errorDialog.js
```diff
@@ -0,0 +1,237 @@
...
+                args: "@expand:fluid.isValue({change}.value)"
+            }
+        }
+    });
+
+    /**
+     * Responsible for the visualisation of the error dialog.
+     */
+    fluid.defaults("gpii.errorDialog", {
+        gradeNames: ["fluid.viewComponent", "gpii.psp.heightObservable"],
```
amb26: Out of interest, what is the cause of the height change (during rendering/loading etc.) that requires this to be observed? The explanation should be added to the comments here

src/main/errorDialog.js
```diff
@@ -0,0 +1,183 @@
...
+        that.events.onContentHeightChanged.fire(height);
+    });
+};
+
+/**
+ * Removes the IPC listeners needed for the communication with the `BrowserWindow`
+ * when the latter is about to be destroyed.
+ */
+gpii.app.errorDialog.channel.deregister = function () {
+    ipcMain.removeAllListeners("onErrorDialogCreated");
```
amb26: I think we've got enough of this type of material now that the design would be substantially rationalised with a generalised IoCSS-based brokering system of the kind you were speculating about a couple of months ago. If we find ourselves without higher priority tasks I think we should go for it - let's talk in the week. This is also something which @the-t-in-rtf started to sketch out a couple of years back.

georgitodorov:
@amb26, I noticed that the test for the width of the survey window started to fail when the electron version got bumped. So I had to remove it. For reference please see the following two Electron issues: https://github.com/electron/electron/issues/10862 and https://github.com/electron/electron/issues/10329.
Also, there seemed to be an issue with destroying webviews in electron 1.8.1. It appears that sometimes the `BrowserWindow` which hosts the webview is destroyed first. Afterwards, when the `webview` is destroyed, it attempts to send a message to the `BrowserWindow` which is no longer there. As a result, we have the following exception when trying to run the tests:
```
FATAL ERROR: Uncaught exception: Object has been destroyed
Error: Object has been destroyed
at WebContents.send (\\vboxsrv\vagrant\node_modules\electron\dist\resources\electron.asar\browser\api\web-contents.js:97:15)
```
You can see the full log here: https://ci.gpii.net/job/gpii-app-test/180/console.

This, however, does not make the tests fail. I tested with the next available stable version of electron (1.8.2) but the issue was also present. In the 2.0.0-beta.1 version the issue is no longer observable. Unfortunately, we do not have much (or any) control of the webview destruction process and I suppose that the only way to fix this is to upgrade to electron 2.0.0 when it becomes stable. Let us know if you have any comments.

src/renderer/errorDialog/js/errorDialog.js
georgitodorov: Added a comment for the grade. Basically, this is necessary because the title, subhead and the actual description of the error are unknown and can have arbitrary length. This on the other hand may necessitate more vertical space than provided in the initial configuration of the `BrowserWindow`. Knowing the exact height makes it possible to adjust the dimensions of the `BrowserWindow` appropriately while not taking any more space than needed.

src/renderer/common/js/heightChangeListener.js
georgitodorov: Actually, the sole purpose of this element is to have listeners attached to it so that changes in its parent's height can be intercepted. I am not sure that the suggested renaming to "contentHeightObserved" will convey this idea better.

src/renderer/common/js/heightChangeListener.js
georgitodorov: This has now been included within the grade's comment.

src/main/psp.js
```diff
@@ -202,10 +300,19 @@ gpii.app.psp.show = function (psp) {
  * changes are `bounds`, `workArea`, `scaleFactor` and `rotation`
  */
 gpii.app.psp.handleDisplayMetricsChange = function (psp, event, display, changedMetrics) {
-    if (changedMetrics.indexOf("workArea") > -1) {
-        var windowSize = psp.pspWindow.getSize(),
-            contentHeight = windowSize[1];
-        psp.resize(contentHeight, true);
+    // In older versions of Electron (e.g. 1.4.1) whenever the DPI was changed, one
```
amb26: Please add a link to the JIRA where this issue is tracked

src/main/psp.js
```diff
@@ -202,10 +300,19 @@ gpii.app.psp.show = function (psp) {
...
+    // `display-metrics-changed` event was fired. In newer versions (e.g. 1.8.1) the
+    // `display-metrics-changed` event is fired multiple times. The change of the DPI
+    // appears to be applied at different times on different machines. On some as soon
+    // as the first `display-metrics-changed` event is fired, the DPI changes are
+    // applied. On others, this is not the case until the event is fired again. That is
+    // why the resizing should happen only the second (or third) time the
+    // `display-metrics-changed` event is fired in which case the changedMetrics argument
+    // will not include the `scaleFactor` string.
+    if (changedMetrics.indexOf("scaleFactor") === -1) {
```
amb26: I think all nodes are sufficiently modern that we can now use the "includes" method https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

src/main/dialog.js
```diff
@@ -119,10 +151,42 @@ fluid.defaults("gpii.app.dialog", {
...
+
+/**
+ * Handle electron's display-metrics-changed event by resizing the dialog if
+ * necessary.
+ * @param that {Component} The `gpii.app.dialog` component.
+ * @param changedMetrics {Array} An array of strings that describe the changes.
+ * Possible changes are `bounds`, `workArea`, `scaleFactor` and `rotation`
+ */
+gpii.app.dialog.handleDisplayMetricsChange = function (that, changedMetrics) {
+    if (changedMetrics.indexOf("scaleFactor") === -1) {
```
amb26: Can now be "includes" (see below)

amb26:
I think this branch is ready to go, with just very minor comments and merging up for conflicts

PR #42, opened 2018-03-28
GPII-2890: PSP does not preserve its aspect ratio when DPI is changed

danailbd:
Improvements over the logic for resizing the window on DPI change (scaling) and on content height changes.


stegru:
ok to test

PR #43, opened 2018-04-02
[WIP] Gpii 2894

danailbd:
### Summary:
* We are using the same mechanism for both the main and the renderer processes that includes:
* A single component that provides messages to all components that require such by using distributeOptions blocks
* This component is backed up by a single “Mega bundle” that is compiled using all available bundles for the different locales. Bundles are searched for and loaded from a given list of directories.
* In the renderer process locale change are propagated via IPC
 
### Details for the PSP i18n architecture decisions:
* Bundles
    * message keys follow the format: `<component_name>_label`
    * support for loading **multiple bundles for the same locale**
 
* Mega Bundle (Bundles compiling)
    * generated from all existing bundle files in the selected directories. It has the following format: `{ <locale>: {<messages>}, en: { <component name>_message1: value, ... }, en_us: {}, bg: {} }`
    * **generated with a Grunt task** which is executed before the application starts
    * different bundles are merged as follows:
        * bundles from all provided folders are collected
        * bundles for each locale are merged sequentially into a single object, i.e. this object contains all messages for the given locale
            * in case of a **key collision**, the value from the last bundle will be used
  	* in case keys are missing for a locale they are searched for sequentially in a less specific locale (if such exists) and finally in the default locale. For example, if we are missing keys in locale `en_us`, a key will first be look for in `en` and then in the default locale. [Ref.](https://github.com/danailbd/gpii-app/blob/GPII-2894/src/common/messageBundlesCompiler.js#L55)
            * order of merging is not guaranteed
            * this algorithm is not currently configurable, as it is done using a Grunt task with a configuration specified in its Gruntfile
 
* Mechanism for messages usage
    * a single grade is added to the root component of a tree which contains i18n-able components
    * this grade is responsible for loading the already generated mega bundle
    * the messages for the current locale are kept in its model( `model: { messages: {} }` ).These messages:
        * are collected from the loaded mega bundle
        * in order to ease distribution, they are **grouped by component name**, e.g. `{ messages: {en: { <component>: { message1: value1, ... } } } }`. Components’ names are extracted from the message keys
        * are updated when the locale changes
    * Distributes messages to interested components:
        * All interested components have a `messages` object in their model which holds the messages used throughout the components
        * once the locale is changed, messages of interested components are updated
        * this update can be intercepted in a model listener in order to make the necessary changes to the UI
        * it is accomplished through the usage of distributeOptions. [Ref.](https://github.com/danailbd/gpii-app/blob/GPII-2894/src/common/messageBundles.js#L201)
    * it uses the fallback mechanism of the bundle compiler
 

#### Possible improvements:
* A dedicated bundle for every component in the format <component name>_locale.json, e.g. gpii-app-menu_en.json, gpii-app-menu_bg.json
    * Keys: Having named the bundles after a component we can easily keep only the label that this message refers to, e.g. for the tray menu, we could simply have a key like - {"keyOutBtn": "Key Out"}


stegru:
ok to test

src/common/messageBundles.js
```diff
@@ -0,0 +1,240 @@
...
+ * @param messageBundles {Object} An object containing all i18n messages for all supported
+ * locales
+ * @param defaultLocale {String} A string representing the default locale.
+ * @return {String} The name of the distributor grade.
+ */
+gpii.app.messageBundles.getMessageDistributorGrade = function (messageBundles, defaultLocale) {
+    var defaultMessages = messageBundles[defaultLocale],
+        distributions = gpii.app.messageBundles.getMessageDistributions(defaultMessages);
+
+    fluid.defaults("gpii.app.messageDistributor", {
```
amb26: This is pretty interesting. On one hand, would prefer this not to be an "app-wide singleton grade" since the scope of an Infusion application might be quite large - what we do in other cases is append a guid to this name to ensure that it can't collide. On the other hand this creates an issue with respect to "garbage collecting" this grade once it is not referenced any more, so some more thought is required here - I'll create a JIRA and link here. In the meantime I think the approach with a guid is preferable.

src/common/messageBundles.js
```diff
@@ -0,0 +1,240 @@
...
+};
+
+/**
+ * Loads synchronously and parses the messageBundles file.
+ * @param messageBundlesPath {String} The path to the messageBundles file relative
+ * to the project directory.
+ * @return {Object} The parsed message bundles for the different locales.
+ */
+gpii.app.messageBundles.loadMessageBundles = function (messageBundlesPath) {
+    var resolvedPath = require("path").resolve(messageBundlesPath);
```
amb26: This should allow module-relative references of the form %module/path and use https://docs.fluidproject.org/infusion/development/NodeAPI.html#fluidmoduleresolvepathpath

amb26: OK, you have explained that because this executes in the renderer process, and also since this only references "built bundles" that the use of fluid.module.resolvePath is impossible/unnecessary

messageBundlesCompiler.js
```diff
@@ -0,0 +1,198 @@
+/**
+ * The message bundles compiler.
+ *
+ * Contains various functions for compiling a mega message bundle which contains
+ * all messages for all available locales.
+ * Copyright 2017 Raising the Floor - International
+ *
+ * Licensed under the New BSD license. You may not use this file except in
+ * compliance with this License.
+ * The research leading to these results has received funding from the European Union's
```
amb26: This section of the boilerplate is no longer needed since the relevant project has ended

messageBundlesCompiler.js
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/**
+ * In case there are only partial messages for a given locale, this function adds
+ * the missing messages using the messages for all versions of the locale in the
+ * order in which they are returned by the `getLocaleVersions` function.
+ * @param messageBundles {Object} A raw map of collected bundles. It is of type
+ * { locale1: { key1: message1, ... }, locale2: { key2: message2 }, ... }
+ * @param defaultLocale {String} The default locale to be used.
+ * @returns {Object} The enhanced message bundles object.
```
amb26: Use a more specific type here such as {MessageBundles} and define its structure in a mega-comment somewhere

src/common/messageBundles.js
```diff
@@ -0,0 +1,240 @@
...
+ */
+gpii.app.messageBundles.getSimpleMessageKey = function (messageKey) {
+    var keyDelimiterIndex = messageKey.lastIndexOf("_");
+    return messageKey.slice(keyDelimiterIndex + 1);
+};
+
+/**
+ * Given a map which contains all messages for a given locale, this function groups the
+ * messages by component grades.
+ * @param {Object} messages A map with all the messages for a given locale.
```
amb26: Use a more specific type than {Object} as per above

src/common/messageBundles.js
```diff
@@ -0,0 +1,240 @@
...
+    var keyDelimiterIndex = messageKey.lastIndexOf("_");
+    return messageKey.slice(0, keyDelimiterIndex);
+};
+
+/**
+ * Given a message key from the `messages` model object, this function returns the
+ * portion of the key which is the simple message key referenced within the corresponding
+ * component. For example, for the "gpii_psp_header_autosaveText" key, this function would
+ * return "autosaveText".
+ * @param messageKey {String} a key from the `messages` object.
```
amb26: This docs comment is the wrong way round - we need to import the new linting tasks into gpii-app - could you please update the package.json to include the most up-to-date versions of the linting tasks as seen in https://github.com/GPII/gpii-express/blob/master/package.json#L23 and relint, thanks (note that the version of eslint-config-fluid with the JSDocs linting tasks is 1.3.0, not 1.2.0 as seen here)

src/common/messageBundles.js
amb26: Can you add a comment linking this to https://issues.fluidproject.org/browse/FLUID-6273 as a place that will be reworked someday and instead make the defaults name as "gpii.app.messageDistributor." + fluid.allocateGuid() - thanks

messageBundlesCompiler.js
```diff
@@ -0,0 +1,198 @@
...
+        });
+
+    return filePaths;
+};
+
+/**
+ * Loads synchronously all message bundles from the passed directories.
+ * @param bundlesDirs {String[]} The list of directrories' names containing message bundles.
+ * @param fileType {String} The file type to be searched for in the given folders.
+ * @param parser {Object} A parser object that has a `parse` method. An example of such is `JSON`.
```
amb26: Rather than accepting a single parser, it would be more helpful to accept a map of file extensions to parsers - e.g. {"json": JSON,
 "json5": JSON5
}

messageBundlesCompiler.js
```diff
@@ -0,0 +1,198 @@
...
+ * compliance with this License.
+ * The research leading to these results has received funding from the European Union's
+ * Seventh Framework Programme (FP7/2007-2013) under grant agreement no. 289016.
+ * You may obtain a copy of the License at
+ * https://github.com/GPII/universal/blob/master/LICENSE.txt
+ */
+
+"use strict";
+
+var fs = require("fs");
```
amb26: I appreciate the worry about doing this from within a grunt task, but I think it would be better in the long run to Infusionise this since it is going to end up in one of @the-t-in-rtf 's packages or else the core framework before long. In practice we found that it was fine to simply issue require("infusion") within these tasks since in any case they cannot execute before npm install has finished, since grunt itself needs to be installed. You can retain the existing style of exporting a map of file exports at the base of the file to ease use within Grunt

PR #45, opened 2018-04-19
GPII-2937: Migration to Electron 1.8.4

danailbd:
In order to use Electron 1.8.4, re-contextualizing for the Electron menu template object is needed.
This is caused by a [change](https://github.com/electron/electron/blob/v1.8.4/lib/browser/api/menu.js#L170) in an Array check to a less graceful one.

Infusion expansion uses a different context which causes this check to fail.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -64,6 +64,11 @@ fluid.defaults("gpii.app.menuInApp", {
   * @param events {Object} An object containing the events that may be fired by items in the menu.
   */
 gpii.app.updateMenu = function (tray, menuTemplate, events) {
+    // Needed in order to get around this non graceful check: https://github.com/electron/electron/blob/v1.8.4/lib/browser/api/menu.js#L170
+    // The infusion's expander uses a context different than the current,
+    // which cases this Array check to fail.
+    menuTemplate = gpii.app.recontextualise(menuTemplate);
+
     menuTemplate = gpii.app.menu.expandMenuTemplate(menuTemplate, events);
 
     tray.setContextMenu(Menu.buildFromTemplate(menuTemplate));
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -64,11 +64,10 @@ fluid.defaults("gpii.app.menuInApp", {
   * @param events {Object} An object containing the events that may be fired by items in the menu.
   */
 gpii.app.updateMenu = function (tray, menuTemplate, events) {
+    // XXX Related to: https://github.com/electron/electron/issues/12698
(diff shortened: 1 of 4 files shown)


src/main/menu.js
```diff
@@ -64,6 +64,11 @@ fluid.defaults("gpii.app.menuInApp", {
   * @param events {Object} An object containing the events that may be fired by items in the menu.
   */
 gpii.app.updateMenu = function (tray, menuTemplate, events) {
+    // Needed in order to get around this non graceful check: https://github.com/electron/electron/blob/v1.8.4/lib/browser/api/menu.js#L170
+    // The infusion's expander uses a context different than the current,
```
javihernandez: ... uses a different context ...

src/main/menu.js
```diff
@@ -64,6 +64,11 @@ fluid.defaults("gpii.app.menuInApp", {
...
+    // which cases this Array check to fail.
```
javihernandez: ... which *causes* ...

src/main/utils.js
```diff
@@ -148,3 +148,21 @@ gpii.app.timer.clear = function (that) {
         that.timer = null;
     }
 };
+
+/**
+ * Set proper context (the current) for elements such as arrays.
+ * This is needed in order for arrays to pass the more strict
+ * check of: `instanceof array`. In generally such checks are to be avoided
```
javihernandez: s/generally/general

javihernandez:
Thanks for this update @danailbd :+1: 
Can you update it according to my comments? In general, I'm fine with the code, although the "recontextualise" function looks like a "workaround" to me, but I'll defer to @amb26, maybe you've already talked about this and found out that this is the only way we can achieve such thing at this moment.

amb26:
It's a workaround but it is actually their bug and not ours, and we should file it with them

javihernandez:
> It's a workaround but it is actually their bug and not ours, and we should file it with them

Sure, I wasn't complaining, just double checking.

src/main/menu.js
amb26: Should be clear that this refers to a V8 JS VM context as documented at https://nodejs.org/api/vm.html

javihernandez:
ok to test

javihernandez:
ok to test

javihernandez:
Tests are passing locally :thinking: 
From the last logs, I'm seeing that the tests are failing right after the "C:\\Windows\\System32\\LicensingUI.exe" process pops up. Also, I'm seeing that CI is using the previous version of our vagrant box (2018.01.25), which has already expired and may be the source of these CI failures.
@amatas, can you nudge our beloved CI to use our last vagrant box?

javihernandez:
ok to test

PR #46, opened 2018-04-23
GPII-2969: Add support for multiple triggers and survey

danailbd:
In order for a different survey to be shown one have to:
- add trigger to the array of triggers in `triggers.json` (or to the `triggers.dev.json` respectively in case using the dev configuration)
- add key with the **id of the trigger** with the desired survey
definition in order for this survey to be shown with the trigger
occurrence


amb26:
Thanks, this looks good in itself - could you update to the current gpii-windows build of 0.3.0-dev.20180625T170252Z.a60081a (identical with current trunk) and check out that it still works in CI - also could you verify that the task tray icon does appear what the app is started - @javihernandez , you mentioned you had found problems with this, are these the ones that are described in https://issues.gpii.net/browse/GPII-2911 or have you/should you file a fresh issue for this?

danailbd:
Generally speaking, we've observed such problems with the tray icon not being shown when there was a `gpii-universal` error before Electron app (`gpii.app`) creation, where Electron code wasn't even reached. Strangely, when I tried the `gpii-app` using the given `gpii-windows` version both the tray icon didn't show up and there weren't any logged errors. On the other hand, when using the original version everything seemed to be working properly. I suppose it’s likely this is a problem with the `gpii-universal` as the `gpii.app` is created as a subcomponent of the `flowManager` which is achieved through a [configuration](https://github.com/GPII/gpii-app/blob/d32a08de0584aa86355d99ab5432f1d8daa72278/configs/app.production.json#L6) that distributes the `gpii.app` there. There might be some unlogged critical error that prevents the creation of the `flowManager` and therefore of the `gpii.app` as well?
With the current dev `gpii-app` version https://github.com/GPII/gpii-app/pull/48 we haven't observed any such problems at least when there isn’t an error logged.
 
As with the latest changes to the surveys in our "QSS 1.1" work _this code is becoming a bit out of date_ I suppose it would be best if we added this code to the “QSS 1.1” codebase either by merging the PR as it is or by closing it and merging the code manually, and further debug this problem in case it appears again…

@amb26 , what do you think?

danailbd:
This will be directly merged with the QSS 1.1 work - #48

PR #47, opened 2018-05-04
Implementation of PSP3 functionalities

georgitodorov:
This PR includes implementation of the following tasks:

- [GPII-2928: Modify nobody keyed in screen](https://issues.gpii.net/browse/GPII-2928)
- [GPII-2929: Add About screen](https://issues.gpii.net/browse/GPII-2929)
- [GPII-2931: Add sound notifications in PSP](https://issues.gpii.net/browse/GPII-2931)
- [GPII-2932: Display image associated with preference set](https://issues.gpii.net/browse/GPII-2932)
- [GPII-2939: Modify PSP system tray icon tooltips](https://issues.gpii.net/browse/GPII-2939)
- [GPII-2940: Grouping functionality in PSP](https://issues.gpii.net/browse/GPII-2940)
- [GPII-2941: Modify PSP application restart behavior](https://issues.gpii.net/browse/GPII-2941)
- [GPII-2988: Implement click-outside user preference](https://issues.gpii.net/browse/GPII-2988) (PSP-side only. When the Core's end of this functionality is ready, it will be pretty straightforward to "merge" them)
- [GPII-2970: Open PSP panel from icon](https://issues.gpii.net/browse/GPII-2970)

PR #48, opened 2018-07-10
Implementation of the QuickSet Strip functionality

georgitodorov:
Some of the code in this PR has been reviewed as part of https://github.com/GPII/gpii-app/pull/47


javihernandez:
ok to test

javihernandez:
ok to test

src/main/psp.js
```diff
@@ -36,12 +37,15 @@ fluid.defaults("gpii.app.pspInApp", {
     model: {
         keyedInUserToken: "{app}.model.keyedInUserToken"
     },
+
     modelListeners: {
-        isShown: {
-            funcName: "gpii.app.hideRestartDialogIfNeeded",
-            args: ["{dialogManager}", "{change}.value"]
+        "{qssWrapper}.qss.model.isShown": {
```
danailbd: This (and all related) will probably not be needed anymore as the PSP can't be shown without the QSS being opened.

src/main/dialogs/basic/dialog.js
```diff
@@ -0,0 +1,433 @@
...
+        setRestrictedSize: {
+            funcName: "gpii.app.dialog.setRestrictedSize",
+            args: [
+                "{that}",
+                "{that}.options.config.restrictions",
+                "{arguments}.0", // width
+                "{arguments}.1"  // height
+            ]
+        },
+        _show: {
```
amb26: showImpl is probably more in line with conventions we use in other areas

src/main/dialogs/basic/dialog.js
```diff
@@ -0,0 +1,433 @@
...
+ * the newly created BrowserWindow
+ * @return {BrowserWindow} The Electron `BrowserWindow` component
+ */
+gpii.app.dialog.makeDialog = function (that, windowOptions, url, params) {
+    var dialog = new BrowserWindow(windowOptions);
+
+    dialog.loadURL(url);
+
+    // Keep record in the window itself for its wrapping grade
+    dialog.grade = that.options.gradeNames[that.options.gradeNames.length - 1];
```
amb26: This is pretty dangerous and will be unstable to options distributions from other parts of the application, etc. 
However I haven't yet found what use is made of this field in any case, we should discuss.

src/main/dialogs/basic/blurrable.js
```diff
@@ -0,0 +1,101 @@
...
+});
+
+/**
+ * Initializes the component by registering a blur listener for the member
+ * dialog.
+ * @param {Component} that - The `gpii.app.blurrable` instance.
+ */
+gpii.app.blurrable.initBlurrable = function (that) {
+    // Attach the grade names as window parameters
+    that.dialog.gradeNames = that.options.gradeNames;
```
amb26: Any reason we can't just use them where they stand in that.options?

src/main/dialogs/basic/dialog.js
amb26: I've found the use of gradeNames for the linkedWindow mechanism, was this intended to be related?

src/main/menu.js
```diff
@@ -92,13 +97,24 @@ fluid.defaults("gpii.app.menuInAppDev", {
             singleTransform: {
                 type: "fluid.transforms.free",
                 func: "gpii.app.menu.generateMenuTemplate",
-                args: ["{that}.model.showPSP", "{that}.model.keyedInSnapset", "{that}.options.locales", "{that}.options.snapsets", "{that}.model.preferenceSetsMenuItems", "{that}.model.keyOut", "{that}.options.exit"]
+                args: [
+                    "{that}.model.showQSS",
```
amb26: This has become really huge and would benefit a lot from being converted into a structured "options" object

src/main/qss.js
```diff
@@ -0,0 +1,620 @@
...
+        }
+    });
+
+    return settings;
+};
+
+/**
+ * Returns all settings (including subsettings) available in the provided
+ * settings groups.
+ * @param {Object[]} settingGroups - An array with setting group items as
```
amb26: Be more  specific about the type here, perhaps via an @typedef

src/main/qss.js
```diff
@@ -0,0 +1,620 @@
...
+    return settings;
+};
+
+/**
+ * When new preferences are delivered to the QSS wrapper, this function takes
+ * care of notifying the QSS about the changes which should in turn update its
+ * internal models and UI. Note that settings changes as a result of a change
+ * in the preference set are not undoable.
+ * @param {Component} that - The `gpii.app.qssWrapper` instance.
+ * @param {Obeject[]} preferences - The new preferences that are delivered to
```
amb26: Typo Object

package.json
```diff
@@ -11,7 +11,8 @@
     "dependencies": {
         "electron": "1.8.4",
```
amb26: Please update to the latest project linting definitions (all other linting dependencies can be removed except for eslint-config-fluid) as seen at https://github.com/fluid-project/infusion/blob/master/package.json#L43 and https://github.com/GPII/universal/blob/master/Gruntfile.js#L18

src/main/dialogs/basic/blurrable.js
georgitodorov: We are using `BrowserWindow.getFocusedWindow()` which returns the currently focused `BrowserWindow` in order to determine if the dialog of the current component should close as a result of a blur somewhere in the component tree. We thought it would be better for the dialog to know about the gradeNames of the component to which belongs instead of to iterate through all components that are dialogs whenever a blur occurs.

package.json
amb26: Please also do a general sweep of outdated dependencies via npm outdated

src/main/dialogs/basic/blurrable.js
amb26: OK, I see - please leave an explanatory comment explaining what is the scope of this usage. Also, write a placeholder comment in the "dialog" member definition of gpii.app.dialog indicating that these grades may be present. If this usage expands further it may be clearer to just inject the entire component or else implement a dedicated id-based lookup from dialogs to components.

package.json
georgitodorov: We changed the linting definitions and cleaned up the newly emerged lint warnings. Most of the dependencies listed in the `package.json` file are related either to the build or testing process and we are not sure whether we are the right people to make these changes. Perhaps @javihernandez and @the-t-in-rtf can help?

src/main/menu.js
georgitodorov: Can you please clarify what you mean?

amb26: I mean sending a structure organised like 
````
{
    showPSP: "{that}.model.showPSP",
    keyedInSnapset: "{that}.model.keyedInSnapset"
````
etc

File: package.json (hunk 8 above)
Comment (the-t-in-rtf) [reply package.json]: @georgitodorov, IMO we should all be comfortable routinely updating the dependencies based on the output of `npm outdated`.  There may be cases where we cannot upgrade beyond a particular point, but in my experience it's better to update often and flag any problems.  Among other things, we pick up security and bug fixes as part of our ongoing work rather than having to charge someone with monitoring and performing the work separately.

File: src/main/dialogs/basic/dialog.js (hunk 3 above)
Comment (danailbd) [reply src/main/dialogs/basic/dialog.js]: Well, this isn’t directly related to the linkedWindows mechanism but it shares the idea. It is used as a way for creating a two-way relation between a dialog and its wrapping component. It looked like a useful feature if we want to get the wrapping component from a BrowserWindow instance. It was meant to use the most specific grade from the list as identifier, but as you proposed, it would probably be best to replace it with a generated unique identifier as generally there might be more than one instance of a single dialog grade.
As an example of its usage you can see in the branch [for the QSS i18n](https://github.com/danailbd/gpii-app/blob/GPII-3169/src/main/dialogs/basic/dialog.js#L315) where this ID is sent from the renderer through the IPC channel to notify its wrapping component about the component’s readiness. We’ll probably be merging this branch tomorrow and I may change it there to make use of a unique identifier. What do you think?

File: package.json (hunk 8 above)
Comment (georgitodorov) [reply package.json]: Okay, I updated the references to the testing-related modules, did some preliminary checks and I think that the testing infrastructure is working fine. As for the `electron-packager`, I see that the module's major version has been bumped by 4 and I am apprehensive to update it without knowing how to check if it is working. That is why my suggestion is to wait for @javihernandez to come back and verify if everything will be correct. As for the update of Electron, we have a separate task for doing this (https://issues.gpii.net/browse/GPII-3044) which may require more work (e.g. more testing to discover if everything behaves correctly). That is why I would propose to leave the update of Electron for another time.

File: src/main/menu.js (hunk 5 above)
Comment (amb26) [reply src/main/menu.js]: Sorry, looking at the implementation now I see its explicit purpose is to process an ordered list, please ignore this comment

File: src/main/dialogs/basic/dialog.js (hunk 3 above)
Comment (georgitodorov) [reply src/main/dialogs/basic/dialog.js]: Removing this for now. An improved version of this mechanism will come in together with the QSS i18n branch.

**Comment by amb26** [comment]
ok to test

**Comment by amb26** [comment]
The test failure on this run looks like a real one - "Error: listen EADDRINUSE :::8058"

**Comment by georgitodorov** [comment]
Interesting - this doesn't happen locally. On my machine tests fail for a different reason ("noUser" does not exist). Also, with the older version of `gpii-windows` tests complete successfully.
Any idea what might be causing this?

File: src/main/dialogs/basic/dialog.js [hunk 9]
```diff
@@ -0,0 +1,433 @@
...
+        setRestrictedSize: {
+            funcName: "gpii.app.dialog.setRestrictedSize",
+            args: [
+                "{that}",
+                "{that}.options.config.restrictions",
+                "{arguments}.0", // width
+                "{arguments}.1"  // height
+            ]
+        },
+        showImp: {
```
amb26: Typo Imp -> Impl

src/main/dialogs/basic/dialog.js
```diff
@@ -0,0 +1,433 @@
...
+            }
+        }
+    },
+    members: {
+        // XXX move to the model probably
+        width:  "{that}.options.config.attrs.width", // the actual width of the content
+        height: "{that}.options.config.attrs.height", // the actual height of the content
+
+        // Blurrable dialogs will have the `gradeNames` property which will contain
+        // all gradeNames of the current component. Useful when performing checks about
```
amb26: Slightly clearer changing current -> containing. Also, make explicit reference to BrowserWindow.getFocusedWindow and the file blurrable.js to help the reader.

amb26:
Closed for https://github.com/GPII/gpii-app/pull/62

PR #59, opened 2018-09-19
GPII-3370: Setting the min/max DPI on QSS.

stegru:
Works well with https://github.com/GPII/windows/pull/199.

Provides the ability to specify dynamic (read at load time) values in the settings schema, so the min/max DPI settings can be used in the screen scale widget.

This is a "better than nothing" approach, as it expects the system's current DPI setting to be at the recommended/default value - otherwise the initial click may not be as expected. Also, changes outside of GPII are not considered - but I believe this to be out of scope for this fix.


stegru:
Moved here: https://github.com/danailbd/gpii-app/pull/45

PR #62, opened 2018-10-08
Implementation of QSS v1.1 functionalities

georgitodorov:


stegru:
ok to test

src/main/common/undoStack.js
```diff
@@ -0,0 +1,125 @@
...
+ * @param {Any} change - The change to be registered.
+ */
+gpii.app.undoStack.registerChange = function (that, change) {
+    var undoStack = fluid.copy(that.model.undoStack);
+
+    undoStack.push(change);
+
+    // get rid of the oldest change if there are a lot of changes
+    if (that.options.maxUndoEntries <= undoStack.length) {
+        undoStack.shift();
```
amb26: This line is untested. We might consider just eliminating the stack size limitation.

src/main/app.js
```diff
@@ -219,217 +373,153 @@ fluid.defaults("gpii.app", {
...
-            args: [
-                "{that}",
-                "{dialogManager}",
-                "{that}.options.errorsDescriptionMap",
-                "{arguments}.0" // error
-            ]
-        }
+gpii.app.changeGpiiAppShortcut = function (shortcutsManager, shortcut, oldShortcut, eventName) {
+    if (oldShortcut) {
+        shortcutsManager.deregisterGlobalShortcut(oldShortcut);
```
amb26: This line is untested

src/shared/messageBundles.js
```diff
@@ -167,24 +165,28 @@ gpii.app.messageBundles.getMessageDistributorGrade = function (messageBundles, d
...
  * to the project directory.
  * @return {Object} The parsed message bundles for the different locales.
  */
 gpii.app.messageBundles.loadMessageBundles = function (messageBundlesPath) {
-    var resolvedPath = require("path").resolve(messageBundlesPath);
+    if (fluid.require) {
+        return fluid.require("%gpii-app/" + messageBundlesPath);
+    }
+    // Required by renderer process
+    var resolvedPath = require("path").join(__dirname, "../../..", messageBundlesPath);
```
amb26: The fact that this line shows up with no coverage suggested we may have some problem with collecting coverage data from electron renderer processes - @georgitodorov , could you verify this is really called and perhaps @the-t-in-rtf  can comment

src/shared/utils.js
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Determines if there is at least one setting group in the provided array of
+ * setting groups which contains at least one setting.
+ * @param {module:gpiiConnector.SettingGroup[]} settingGroups - An array of
+ * setting groups.
+ * @return {Boolean} `true` - if there is at least one setting group with at
+ * least one setting in it.
+ */
+gpii.app.settingGroups.hasSettings = function (settingGroups) {
```
amb26: This function is unused

src/shared/messageBundles.js
georgitodorov: Right, currently the coverage data is only for the main process.

src/shared/utils.js
georgitodorov: I guess, you mean untested? Because it is actually used here (even though this is yet another untested piece of code): https://github.com/danailbd/gpii-app/blob/feature/QSS1.1/src/main/surveys/surveyTriggerManager.js#L417.

This brings up another thing - in order to test some of the PSP/QSS functionalities, it will be good if we can have a profile without any settings. Is there such a user profile? If not, can one be created or can we mock such if not?

amb26: I don't believe there is such a profile, please issue a small pull to universal if you would like

amb26: You just need to do that, but also remember to rerun the dbData build process which is attached to npm install

src/main/shortcutsManager.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    globalShortcut.register(command, shortcutEvent.fire);
+};
+
+/**
+ * Deregisters a global shortcut.
+ * @param {String} command - The global shortcut string.
+ */
+gpii.app.shortcutsManager.deregisterGlobalShortcut = function (command) {
```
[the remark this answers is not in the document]
georgitodorov: Just added various tests for the `ShortcutsManager`.

the-t-in-rtf:
@georgitodorov, I finished my troubleshooting and narrowed down the hanging behavior to the qss tests here:

https://github.com/danailbd/gpii-app/blob/feature/QSS1.1/tests/IntegrationTests.js#L114

To reproduce in a vagrant VM:

1. cd \vagrant
2. npm install
3. npm run test:psp
4. attempt to close the shell after using the "exit" command.

In my testing, only the integration tests result in a hung shell, and of those, only the qss tests result in a hung shell.

amb26:
ok to test

javihernandez:
ok to test

src/main/app.js
georgitodorov: I have just added some additional tests which cover this.

danailbd:
I've been running some diagnostics of the CI tests stability by amending the last commit and it behaves really strange. Currently there are 2 different failures so I will continue debugging it.

PR #68, opened 2018-11-27
GPII-3550: Converted messageBundlesCompiler.js into a script

javihernandez:
Here's a first implementation of this. There is a lot of room for improvements, so let me know your thoughts on this.

### DIFF
--- a/scripts/compileMessageBundles.js
+++ b/scripts/compileMessageBundles.js
@@ -1,0 +1,57 @@
+/**
+ * The message bundles compiler script.
+ *
+ * This script compiles a mega message bundle which contains
+ * all messages for all available locales.
+ * Copyright 2018 Raising the Floor - International
+ *
+ * Licensed under the New BSD license. You may not use this file except in
+ * compliance with this License.
+ * The research leading to these results has received funding from the European Union's
+ * Seventh Framework Programme (FP7/2007-2013) under grant agreement no. 289016.
+ * You may obtain a copy of the License at
+ * https://github.com/GPII/universal/blob/master/LICENSE.txt
+ */
+
+"use strict";
+
+var fluid = require("infusion");
+var gpii = fluid.registerNamespace("gpii");
+
+var fs = require("fs");
+var path = require("path");
+var shell = require("shelljs");
+
+require("./shared/messageBundlesCompiler.js");
+
+/**
+ * This function generates the messages bundle and writes into a given file.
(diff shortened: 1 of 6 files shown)


danailbd:
It looks good. My only concern is with the [tests for the compiler](https://github.com/GPII/gpii-app/blob/master/tests/MessageBundlesCompilerTests.js). The `messageBundlesCompiler.js` is simply loaded which would cause the script to execute creating the `build` folder etc.
We could extract the script for execution in a separate js file.

danailbd:
> CI job failed: https://ci.gpii.net/job/gpii-app-tests/441/

Well, that looks strange: `Error: Cannot find module '../scripts/shared/messageBundlesCompiler.js'`

tests/MessageBundlesCompilerTests.js
```diff
@@ -19,7 +19,7 @@ var gpii = fluid.registerNamespace("gpii");
 
 var jqUnit = fluid.require("node-jqunit", require, "jqUnit");
 
-require("../messageBundlesCompiler.js");
+require("../scripts/shared/messageBundlesCompiler.js");
```
danailbd: Changing to `fluid.require("%gpii-app/scripts/shared/messageBundlesCompiler")` might help

javihernandez: Nice catch, I didn't realize because I ran the affected test only (without coverage reporting). I've updated it following your suggestion.

scripts/compileMessageBundles.js
```diff
@@ -0,0 +1,57 @@
...
+require("./shared/messageBundlesCompiler.js");
+
+/**
+ * This function generates the messages bundle and writes into a given file.
+ * @param {String[]} messageDirs - An array of the directories from which messages
+ * files are to be retrieved.
+ * @param {String}  resultFilePath - The file where the bundles are going to be written.
+ */
+gpii.app.compileMessageBundles = function (messageDirs, resultFilePath) {
+    require("gpii-windows");
```
danailbd: We should also use `require("gpii-windows/index.js")` as `require("gpis-windows")` also runs gpii (see `gpii-windows/gpii.js`)
I just spent some time struggling with this...

javihernandez: Thanks for your testing @danailbd. I've just added that change.

danailbd:
I think we can go ahead and merge this PR

PR #69, opened 2018-11-28
Fix app dependencies

danailbd:
Ensure that `gpii-app` is requiring modules that it explicitly depends on:
- if `gpii-universal` is needed - use `gpii-windows` to load it
- include other missing dependencies - `ws`

Also restore installer creation which was stopped to improve CI tests debugging speed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
         "infusion": "3.0.0-dev.20180222T160835Z.6e1311a",
         "node-jqunit": "1.1.8",
         "request": "2.88.0",
-        "winstrap": "0.5.12"
+        "winstrap": "0.5.12",
+        "ws": "6.1.2"
     },
     "devDependencies": {
         "electron-packager": "8.5.1",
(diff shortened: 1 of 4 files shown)


javihernandez:
ok to test

package.json
```diff
@@ -16,7 +16,8 @@
         "infusion": "3.0.0-dev.20180222T160835Z.6e1311a",
         "node-jqunit": "1.1.8",
         "request": "2.88.0",
-        "winstrap": "0.5.12"
+        "winstrap": "0.5.12",
+        "ws": "6.1.2"
```
javihernandez: I'm curious, where is this dependency coming from?

javihernandez: Oh, nevermind, I'm just seeing that we're using it in the gpiiConnector.

javihernandez: @amb26 I see that kettle is using 6.0.0 while we're adding 6.1.2 in this pull request. I don't see any problem with that, but just wanted to double check with you in case you see any potential issue or if we would like to use the same version in both sides.

amb26: Hi @javihernandez - this is fine - ws artefacts never get shared across the module boundary. In any case Kettle is tested fine with 6.1.x in this pull in progress - https://github.com/fluid-project/kettle/pull/49

javihernandez: okay, thanks for confirming

PR #70, opened 2018-11-30
GPII-3540: Implemented the tray button.

stegru:
Here's the tray button.

The button (this thing) and icon (the usual thing) can be enabled using the `trayType` field in the siteconfig file. Currently, both are turned on.

Requires https://github.com/GPII/windows/pull/222

To test: It should work exactly the same as the notification icon.


stegru:
only gpii-windows failures, that are now fixed

ok to test

provisioning/Build.ps1
```diff
@@ -61,6 +61,9 @@ If ($mainDir -eq "C:\vagrant") {
     $npmInstallDir = (get-item $originalBuildScriptPath).parent.FullName
 }
 
+# Build the tray button
+../trayButton/build.ps1
```
javihernandez: I'm wondering whether we should build the native tray button as a postinstall action rather than from a provisioning script. Same applies for generating the language bundle. @stegru @amb26 @danailbd what do you think?

javihernandez: Yes, I think it's fine, you can just call the build script from the npm install (or postinstall) script like this:

`"postinstall": "npm run build"`

and leave the "build" npm script as you have it already (compiles messages and build the native tray).
What I want to avoid here is to end up without the tray after running npm install (we don't always run the provisioning scripts)

stegru:
failure of gpii-windows
ok to test

package.json
```diff
@@ -12,13 +12,12 @@
...
         "electron-localshortcut": "3.1.0",
-        "gpii-windows": "0.3.0-dev.20181120T145240Z.c46b4be",
+        "gpii-windows": "stegru/windows#GPII-3540",
         "infusion": "3.0.0-dev.20180222T160835Z.6e1311a",
         "node-jqunit": "1.1.8",
         "request": "2.88.0",
         "winstrap": "0.5.12",
-        "ws": "6.1.2"
-    },
+        "ws": "6.1.2"},
```
javihernandez: Extra character: }

stegru:
ok to test

stegru:
ok to test

stegru:
ok to test

stegru:
ok to test

javihernandez:
I guess that this is ok to test again

src/main/tray.js
```diff
@@ -174,3 +139,239 @@ gpii.app.getTrayTooltip = function (isKeyedIn, preferences, messages) {
...
+        },
+        isKeyedIn: {
+            func: "{that}.updateButton",
+            args: [ "{that}.buttonItems.state", "{change}.value" ]
+        }
+    },
+    model: {
+        highContrastIcon: "{that}.options.icons.highContrastIcon"
+    },
+    members: {
```
amb26: No need for any non-modifiable material to be members.

package.json
```diff
@@ -12,7 +12,7 @@
         "electron": "3.0.2",
         "electron-edge-js": "8.3.8",
         "electron-localshortcut": "3.1.0",
-        "gpii-windows": "0.3.0-dev.20181213T111654Z.3af6c04",
+        "gpii-windows": "stegru/windows#GPII-3540",
```
amb26: Should be able to update to 0.3.0-dev.20190111T181517Z.da9a4e1

PR #71, opened 2018-12-10
GPII-3455: Fixing reset to standard via desktop link

stegru:
Fixes the following error, when the reset link is used:

```
Reset: Error occurred during logout: Unable to cast COM object of type 'System.__ComObject' to interface type 'System.Management.IWbemServices'. This operation failed because the QueryInterface call on the COM component for the interface with IID '{9556DC99-828C-11CF-A37E-00AA003240C7}' failed due to the following error: The application called an interface that was marshalled for a different thread. (Exception from HRESULT: 0x8001010E (RPC_E_WRONG_THREAD)).
```

Calling via `nextTick` used to work, but now it doesn't (Electron 3 upgrade?). Instead, calling via `setTimeout(..., 0)` works.

I don't know why this fixes it, nor do I know what's causing it.

Perhaps it's due to electron (or chrome) also using the COM object. or maybe there's an instance in `dotNetProcesses.csx` that isn't being disposed, but I'm not sure why it's complaining about the wrong thread.

The only thing I do know, is it now works. But I'm sceptical about that, too.


stegru:
oh, also I've used the new way of ensuring a single instance as the previous method is deprecated.

https://electronjs.org/docs/all#appmakesingleinstance

main.js
```diff
@@ -33,25 +43,18 @@ app.disableHardwareAcceleration();
 // The PSP will have a single instance. If an attempt to start a second instance is made,
 // the second one will be closed and the callback provided to `app.makeSingleInstance`
 // in the first instance will be triggered enabling it to show the PSP `BrowserWindow`.
-var appIsRunning = app.makeSingleInstance(function (commandLine) {
+app.on("second-instance", function (event, commandLine) {
```
javihernandez: Question. Should this event listener be set up from the "gpii.app" component at the initialization process? 
@danailbd ?

javihernandez:
Yes, this seems to fix the problem here.

main.js
```diff
@@ -33,25 +43,18 @@ app.disableHardwareAcceleration();
...
     var qssWrapper = fluid.queryIoCSelector(fluid.rootComponent, "gpii.app.qssWrapper")[0];
     qssWrapper.qss.show();
-
     if (commandLine.indexOf("--reset") > -1) {
```
stegru: This is the main fix here. The above replacement of `makeSingleInstance` with `app.on("second-instance")` makes no difference to the problem, but it's deprecated in electron 3 (and removed in 4)

main.js
```diff
@@ -33,25 +43,18 @@ app.disableHardwareAcceleration();
...
-        process.nextTick(function () {
+        setTimeout(function () {
             // GPII-3455: Call this in the next tick, to allow electron to free some things.
```
amb26: Please adjust this comment to reflect the fact that we seem to know that, for some reason, process.nextTick doesn't work any more (got even any vague guesses about why it doesn't?)

stegru: Perhaps it has something to do with the way this callback is invoked. I looked at the electron code, and there's nothing obvious - the new process just sends a window message to the existing one, which raises the event. I've checked it's the same thread. Perhaps something is/isn't initialised above this invocation.

As to why it no longer works since Electron 3, perhaps the implementation of nextTick has changed slightly. Or maybe it's related to a change in edge.js. I just don't know.

I don't believe it's a time/delay issue - otherwise a zero delay wouldn't always work. But, if that turns out to be the case then we'll look into another way of implementing this rather than continue to chase a magic timing sweet spot.

PR #75, opened 2018-12-15
GPII-3597: Replaced the SVG spinner with an animated GIF

javihernandez:
This change reduces the CPU load considerably and it doesn't keep raising anymore.


javihernandez:
ok to test

danailbd:
I've tried it and I also observe the CPU load being about 2 or 3 times less compared to the SVG approach. For a virtual machine with 4 cores it's between 1-2% load.

javihernandez:
Closed in favor of https://github.com/GPII/gpii-app/pull/78

PR #76, opened 2018-12-19
GPII-3597: Only animating the spinner when it's shown.

stegru:
Stops burning the CPU on animating something that isn't visible.


src/renderer/waitDialog/css/message.css
```diff
@@ -31,7 +31,7 @@
     padding: .5em;
 }
 
-.spinner {
+.animate .spinner {
```
danailbd: We have followed a convention where the "action" class placed last. Here we'd use `.spinner .animate` instead.

src/main/dialogs/waitDialog.js
```diff
@@ -118,8 +119,17 @@ gpii.app.waitDialog.hide = function (that) {
...
         that.dialog.hide();
+        gpii.app.waitDialog.toggleAnimation(that, false);
     }
 };
+
+gpii.app.waitDialog.toggleAnimation = function (that, animate) {
+    var script = fluid.stringTemplate("jQuery(document.body).toggleClass(\"animate\", %animate)", {
+        animate: animate
+    });
+    that.dialog.webContents.executeJavaScript(script);
```
danailbd: We've tried to keep the main and the renderer processes as decoupled as possible using IPC and leaving changes to the DOM  to be done by the renderer process but here it's probably a reasonable exception as the `waitDialog` doesn't have a `.js` file on its own.

stegru:
Because the CPU normalises after the spinner is shown, I'm considering a work-around of showing the animation during start-up.

danailbd:
> Because the CPU normalises after the spinner is shown, I'm considering a work-around of showing the animation during start-up.

Sounds like a reasonable solution, but we might want to avoid the user seeing the spinner. We could move the spinner offscreen for the initial show using [moveOffScreen](https://github.com/danailbd/gpii-app/blob/master/src/main/dialogs/basic/offScreenHidable.js#L93) followed by a `setPosition` and normal hide once the spinner has loaded.

src/renderer/waitDialog/css/message.css
stegru: I add the `animate` class to the body, so `.animate` is a parent of `.spinner`.

(I could have added the class to the spinner, but that would increase the coupling)

src/main/dialogs/waitDialog.js
stegru: Yes, I completely agree with you. I tried to keep it all in the renderer, however I couldn't find a way to detect if the dialog had been opened.

stegru:
I was thinking about showing the animation while morphic was loading - now that I've thought about it, it's not a good idea.

PR #77, opened 2018-12-21
Morphic 1.1

danailbd:
#### Includes tasks:
* GPII-3546: New look to the QuickStrip
* GPII-3521: Folding of the QuickStrip when it gets to large (hiding the logo of the new designs)
* GPII-3574: Read Aloud is a Menu
* GPII-3575: Direct control dots in Screen Zoom menu
* GPII-3573: New keyboard navigation for QuickStrip
* GPII-3176: Expand the QSS menu to fit in all available setting values
* GPII-3608: The QS "Screen Zoom" and the PSP "DPI Scale factor" settings' ranges differ
  * A temporary fix on gpii-app side for both "Scale factor" and "Language"
* GPII-3556: Show the survey without focus and behind the QS
* GPII-3557: Update the tooltip for "Reset to Standard" button
* GPII-3662: Language setting in the PSP doesn't reflect OS installed languages
* _(partly)_ GPII-3664: Zooming in and out should not work in any BrowserWindow
  * **(not done)** Redirect zoom in/out inside the QS to the topmost window (similar to "App Text Zoom" setting)
* GPII-3688: Close the Sign-in menu on second click of the button
* GPII-3689: Modify the tooltips of the QS buttons
* GPII-3690: Update the visual behavior of the Screen Zoom menu
* GPII-3691: Modify the behavior of the QS when screen zoom is modified


javihernandez:
ok to test

tests/QssTestDefs.js
```diff
@@ -22,37 +22,68 @@ var fluid = require("infusion"),
...
     clickIncreaseBtn = "jQuery('.flc-qssStepperWidget-incBtn').click()",
-    clickDecreaseBtn = "jQuery('.flc-qssStepperWidget-decBtn').click()";
+    clickDecreaseBtn = "jQuery('.flc-qssStepperWidget-decBtn').click()",
+    clickToggleBtn = "jQuery('.flc-switchUI-control').click()";
 
 // Generic
 var closeClosableDialog = "jQuery(\".flc-closeBtn\").click()";
 
+var openReadAloudMenuSeqEl = {
+    task: "gpii.test.executeJavaScript",
```
amb26: This function has an unhelpfully generic name. Perhaps it would more accurately be something like gpii.test.executeWebContentsJavaScript? Even this doesn't seem quite right since it doesn't execute arbitrary code. Maybe gpii.test.invokeFunctionInWebContents?

src/renderer/qssWidget/js/qssStepperWidget.js
```diff
@@ -346,4 +356,149 @@
...
+     */
+    gpii.qssWidget.stepper.getIndicatorsList = function (setting) {
+        if (!Number.isInteger(setting.schema.min) || !Number.isInteger(setting.schema.max)) {
+            return [];
+        }
+
+        var indicators = [];
+
+        for (
+            var indicatorValue = setting.schema.min ;
```
amb26: Odd, does this semicolon pass linting?

danailbd: Hmm, strangely it does :/

src/renderer/qss/js/qssFocusManager.js
```diff
@@ -0,0 +1,343 @@
...
+     * Focuses the first available button which conforms to all of the following conditions:
+     * 1. The button is focusable.
+     * 2. The button is in the same focus group as the currently focused button.
+     * 3. The button is the first button before or after the currently focused button (depending
+     * on the `backwards` argument) which conforms to the two conditions above.
+     * @param {Component} that - The `gpii.qss.qssFocusManager` instance.
+     * @param {Boolean} backwards - If `true` the scanning direction will be from bottom to top.
+     * Otherwise, it will be from top to bottom.
+     */
+    gpii.qss.qssFocusManager.focusNearestVertically = function (that, backwards) {
```
amb26: It is these which should be the exposed operations on the component, rather than being just used in the impl - people bringing alternative interactions/a11y technologies to the component will want to talk in terms of the focus operation rather than the name of the keypress.

src/renderer/qss/js/qssFocusManager.js
```diff
@@ -0,0 +1,343 @@
...
+
+        maxElementsPerFocusGroup: 2,
+
+        styles: {
+            button: "fl-qss-button",
+            smallButton: "fl-qss-smallButton"
+        },
+
+        listeners: {
+            "onArrowUpPressed.impl": {
```
amb26: These should be rewritten to visibly make calls to the focusedNearestHorizontally etc. methods

tests/QssTestDefs.js
danailbd: Yes, it really sounds too generic. We were following the naming of the Electron WebContents api.
Well, currently this method simply executes a js as a string so I think the first name fits better

On the other hand it seems more convenient to use functions instead of a script strings. Lately, we've been using a "functions approach", e.g. [here](https://github.com/GPII/gpii-app/pull/77/files#diff-320839322d3d294a4886a140356266feR139), so I can go ahead and add such a utility function

src/main/dialogs/basic/dialog.js
```diff
@@ -345,6 +391,96 @@ gpii.app.dialog.positionOnInit = function (that) {
...
+ */
+gpii.app.dialog.rescaleDialog = function (that, scaleFactor, oldScaleFactor) {
+    scaleFactor = scaleFactor || 1;
+    oldScaleFactor = oldScaleFactor || 1;
+
+    var width = that.getScaledWidth(scaleFactor, oldScaleFactor),
+        height = that.getScaledHeight(scaleFactor, oldScaleFactor),
+        offset = that.getScaledOffset(scaleFactor, oldScaleFactor);
+
+    that.applier.change("width", width);
```
amb26: It might help to avoid some kinds of jank if all these changes are made simultaneously?
Either that.applier.change("", {width: width, height: height, offset: offset}) or else wrap this in a transaction

src/main/dialogs/basic/dialog.js
```diff
@@ -183,13 +194,48 @@ fluid.defaults("gpii.app.dialog", {
...
+            args: ["{that}", "{that}.model.scaleFactor"]
+        },
         "onDestroy.cleanupElectron": {
             this: "{that}.dialog",
             method: "destroy",
             priority: "last"
         }
     },
     invokers: {
+        getScaledWidth: {
```
amb26: Rather than 3 invokers, might be neater to just put the scaled variants of these values into the model alongside the unscaled ones using a relay, then they can just be read like other values.

tests/QssTestDefs.js
amb26: Yeah, that sounds good. Something with the signature of and similar name to https://docs.fluidproject.org/infusion/development/CoreAPI.html#fluidinvokeglobalfunctionfunctionpath-args , only acting in the webContents, would be a useful API

tests/testUtils.js
```diff
@@ -148,10 +148,48 @@ gpii.tests.blockTestsElement = function () {
...
  */
-gpii.test.executeJavaScript = function (dialog, command) {
+gpii.test.executeJavaScriptInWebContents = function (dialog, command) {
     return dialog.webContents.executeJavaScript(command, true);
 };
 
+
+/**
+ * Invokes a JavaScript function in the `BrowserWindow` of the given dialog.
+ * It simply transforms the function and passes it to `gpii.test.executeJavaScriptInWebContents`
```
amb26: Explain what transform is done

tests/testUtils.js
```diff
@@ -148,10 +148,48 @@ gpii.tests.blockTestsElement = function () {
...
+    return gpii.test.executeJavaScriptInWebContents(dialog, gpii.test.toIIFEString(func));
+};
+
+/**
+ * Invokes a JavaScript function in the `BrowserWindow` of the given dialog.
+ * Similar to `executeJavaScriptInWebContentsDelayed` invokes the function with a delay.
+ * @param {BrowserWindow} dialog - The `BrowserWindow` in which the script is
+ * to be executed.
+ * @param {Function} func - The function that is to be invoked inside the BrowserWindow
+ * @param {Number} delay - The delay after which the function should be invoked
```
amb26: Presumably delay in ms

src/renderer/qssWidget/js/qssWidget.js
```diff
@@ -228,43 +248,53 @@
...
+     * @param {Component} that - The `gpii.psp.qssWidget` instance.
+     * @param {String} widgetGrade - A grade name for the widget component.
+     * @return {jQuery} The jQuery element representing the element in the DOM or
+     * `undefined` if there is no such element.
      */
-    gpii.psp.qssWidget.getWidgetType = function (setting) {
-        return setting.schema.type === "number" ? "gpii.qssWidget.stepper" : "gpii.qssWidget.menu";
+    gpii.psp.qssWidget.locateDomElement = function (that, widgetGrade) {
+        if (widgetGrade) {
+            var lastDotIndex = widgetGrade.lastIndexOf("."),
```
amb26: Package these two lines as a reusable "widgetGradeToSelectorName" utility

src/renderer/qss/js/qssFocusManager.js
```diff
@@ -0,0 +1,343 @@
...
+                args: ["{that}"]
+            },
+            "onArrowRightPressed.impl": {
+                funcName: "gpii.qss.qssFocusManager.onArrowRightPressed",
+                args: ["{that}"]
+            }
+        },
+
+        invokers: {
+            getFocusInfo: {
```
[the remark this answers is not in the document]
danailbd: @amb26, I suggest that we keep the current implementation for the Morphic 1.1 release. It has been tested thoroughly which has shown its stability. Also, the FocusManager has already been part of master for some time, so if we decide to merge PR 1.1 to master, this would not introduce the functionality, but rather an expansion to its current state. 
I’ve almost completed the desired refactoring but as this is a core functionality, I’m concerned that there might be some cases that I’m missing.
As there are higher priority items, I think it would be better if we address those first.
I’ll open a separate PR for the refactoring addressing all of the comments related to the `focusManager`, and in case I don’t have the time to finalise it, someone else, probably HyperAspect, can resume working on it at a later stage.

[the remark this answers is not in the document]
amb26: No worries, @danailbd - I've written up this requirement under https://issues.gpii.net/browse/GPII-3685 so it can be referred to in future work. I agree we can go ahead merging this pull and table this for work at a future date.

tests/QssTestDefs.js
danailbd: I also put some thought in a way to **provide arguments to the invoked function** but it's not that straightforward as the function code is sent to the renderer as a string.
 
We could, do something like:
```
function argumentsToString (args) {
    var varsSrting = "",
        varTemplate = "%name=%value;";
    fluid.each(args, function (value, name) {
        varsSrting += fluid.stringTemplate(varTemplate, {name: name, value: value});
    });
 
    return varsSrting;
};

gpii.test.toIIFEString = function (func, args) {
     return fluid.stringTemplate("(%func)(%args)"), { func: func.toString(), args: fluid.keys(args).join(",").toString();
}
 
gpii.test.invokeFunctionInWebContents = function (dialog, func, args) {
    var varsString = argumentsToString (args),
        command = fluid.stringTemplate("%arguments;%func", {
            func: gpii.test.toIIFEString(func, args), // adds arg keys to the function invoke string
            arguments: varsSrting
        });
    return dialog.webContents.executeJavaScriptInWebContents(command, true);
};
 
gpii.test.invokeFunctionInWebContents(dialog, someFuncName, { arg1: "value1", arg2: 2 })
```
This will add variable declarations at the beginning of the code string which are also added to the IIFE part of the string.
This should work and would add some flexibility but I have mixed feelings about it...

amb26: I think this could be simplified by just embedding the argument values directly within the function invocation, especially since we don't want to dump a lot of extra conflicting definitions into what I imagine will be the global scope of the browser

danailbd: Yep, that sounds more reasonable. I've come up with this idea from a different angle and totally missed this possibility. So it would look something like:
```
gpii.test.toIIFEString = function (func, args) {
     return fluid.stringTemplate("(%func)(%args)"), { 
                  func: func.toString(),
                  args: args.join(",").toString(); // I think special treatment would be needed for string type - probably addition of quotes
} 
```

siteconfig.json5
```diff
@@ -20,10 +23,12 @@
...
             genericLanguage: "%native · %local"
         },
-        // The code of the language that would always stay on top of the "Language" setting's menu list.
+        // The code of OS default language `gpii-app` (it might be the one with which the system was installed with).
+        // This language should always stay on top of the "Language" setting's menu list.
         // The code must be exactly matching one of the installed languages' code (with respect to the region).
         // In case the selected language code is missing, there simply won't be a special first setting.
-        defaultLanguage: "en-US"
+        // NOTE: this is not related to the default language value defined for the language setting
+        systemDefaultLanguage: "bg-BG"
```
amb26: Did this get left behind from testing?

src/main/dialogs/basic/dialog.js
amb26: This comment still outstanding

siteconfig.json5
danailbd: Yes, I think that's the case. It should be reverted back to "en-US"

src/main/dialogs/quickSetStrip/qssDialog.js
```diff
@@ -153,10 +171,111 @@ fluid.defaults("gpii.app.qss", {
...
+
+
+/**
+ * Computes the total width of all of the QSS buttons, based on their sizes inside
+ * the BrowserWindow.
+ * @param {Component} that - The `gpii.app.qss` instance
+ * @param {Object[]} buttons - The list of QSS buttons
+ * @return {Number} - The total scaled size of the QSS's buttons
+ */
+gpii.app.qss.computeQssButtonsWidth = function (that, buttons) {
```
amb26: Would be clearer here to pass in that.options and that.model separately rather than the whole component, to emphasise that this is a pure calculation function that doesn't have side effects on the component.

src/main/dialogs/quickSetStrip/qssDialog.js
```diff
@@ -153,10 +171,111 @@ fluid.defaults("gpii.app.qss", {
...
     }
 });
 
+
+
+/**
+ * Computes the total width of all of the QSS buttons, based on their sizes inside
+ * the BrowserWindow.
+ * @param {Component} that - The `gpii.app.qss` instance
+ * @param {Object[]} buttons - The list of QSS buttons
```
amb26: Be more specific about the type here - use a @typedef if appropriate

src/main/dialogs/basic/dialog.js
danailbd: In the current model we already keep the scaled variants of the metrics (that’s the reason for using the old `scaleFactor` when calculating the rescaled metrics). These three methods simply recompute the metrics on a change in the scaling factor… so probably a name such as `getRescaled<Metric>` would suit them better.
 
The simplest solutions I can think of that would fulfill your comment are:
* In order to have the scaled versions together with the unscaled ones we might simply compute the unscaled versions based on the updates of the currently existing scaled properties (width, height and offset). It would further follow your suggestion. However, I would prefer a different approach.
* On the other hand, without the unscaled versions we might simply have a `modelRelay` that watches for changes in the `scaleFactor` and updates the metrics accordingly. In combination with a `modelListener` for the `scaleFactor` (assuming the `modelRelay` executes first) that applies the newly computed metrics and updates the zooming of the corresponding BrowserWindow this should do it.  

How do these two sound to you, am I following your idea correctly?
 
Alternatively, the overall scaling mechanism could be re-engineered so that we make use of both the scaled and unscaled versions of the metrics.

amb26: OK, now I understand properly that this is an instantaneous, historyful interaction I think we should not try to involve the model system in this. I suggest instead we just consolidate these three invokers into a single one which accepts a {width:, height:, offset} structure (thus avoiding wholesale dependence on "that") and returns a scaled version of it, given these appear to always be requested simultaneously.

PR #79, opened 2019-01-07
GPII-3607: Use a safer mechanism when moving a dialog from off-screen

danailbd:
``` As of 01.07.2019 PSP status meeting, this should be included in Morphic version 1.1```

Somewhat partial fix for GPII-3607.
It is partial as in case there's a display bellow the primary display that has `scaleFactor` different from the `scaleFactor` of the primary display, it will be positioned and sized correctly, but there also will be a flicker of QS dialogs.

Detailed description of the solution can be found in the [Jira ticket](https://issues.gpii.net/browse/GPII-3607?focusedCommentId=36601&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-36601)


javihernandez:
ok to test

PR #83, opened 2019-01-18
GPII-1908 Initial Prototype of Capture Tool

sgithens:
This is an initial work in progress for review.  I don't expect this current code to be merged.


sgithens:
@amb26 I'm looking in to the failures on CI, but in the meantime could you give this an initial review.

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+
+require("../../../node_modules/gpii-universal/gpii/node_modules/solutionsRegistry/src/js/SolutionsUtils.js");
+
+fluid.defaults("gpii.app.captureToolUtils", {
+    gradeNames: ["fluid.component"],
+    events: {
+        onCollectDiagnostics: null
+    },
+    listeners: {
+        onCollectDiagnostics: [{
```
amb26: These should be namespaced and sequenced using after:xxxx priorities. The framework has no particular reason to enforce the sequence in this array.

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+require("./basic/centeredDialog.js");
+require("./basic/resizable.js");
+require("./basic/offScreenHidable.js");
+
+require("../../../node_modules/gpii-universal/gpii/node_modules/solutionsRegistry/src/js/SolutionsUtils.js");
+
+fluid.defaults("gpii.app.captureToolUtils", {
+    gradeNames: ["fluid.component"],
+    events: {
+        onCollectDiagnostics: null
```
amb26: Explain that this is a pseudoevent and give a clue where it is operated from. Preferably give this component a meaningful name such as "diagnosticsCollector"

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+ * + gpiiSettingsDir/
+ *   + capture-diagnostics-TIMESTAMP
+ *     - solutions.json
+ *     - installedSolutions.json
+ *     - settingsCapture.json
+ *
+ * @param {gpii.app.captureToolUtils} captureToolUtils = An instance of `gpii.app.captureToolUtils`
+ */
+gpii.app.captureTool.logCaptureDiagnostics = function (captureToolUtils) {
+    var settingsDirComponent = gpii.settingsDir();
```
amb26: Don't create a free component like this, it will leak garbage and create resolution hazards. Instead inject a reference to the existing one. I know this pattern exists in another file but it has been rewritten in my upcoming branch.

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+    var gpiiSettingsDir = settingsDirComponent.getGpiiSettingsDir();
+
+    var startupTime = Date.now();
+    var captureDirName = gpiiSettingsDir + "/capture-diagnostics-" + gpii.journal.formatTimestamp(startupTime);
+
+    fluid.log("Capture Diagnostics directory is: ", captureDirName);
+
+    var dailogPromise = electron.dialog.showMessageBox({
```
amb26: typo dialog

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+        buttons: ["Cancel", "Run Capture"],
+        message: "Select 'Run Capture' to run the capture diagnostics and store them to disk.\n\n  After capture they will be stored in:\n" + captureDirName
+    });
+
+    if (dailogPromise === 1) {
```
amb26: What's the significance of the number 1? How could a promise take on the value 1?

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+                fluid.each(data, function (solutionEntry, solutionId) {
+                    try {
+                        fluid.log("ABOUT TO CHECK PROCESS FOR: ", solutionId);
+                        var isRunningState = gpii.processReporter.handleIsRunning(solutionEntry);
+                        fluid.log("RUNNING STATE: ", isRunningState);
+                        if (isRunningState) {
+                            runningSolutions[solutionId] = solutionEntry;
+                        }
+                    } catch (err) {
+                        fluid.log("Exception trying to look up", solutionEntry.isRunning);
```
amb26: Unlikely that this message will be meaningful. At least log the exception somehow, and also check whether "isRunning" is even an existing string member

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+                });
+                event.sender.send("sendingRunningSolutions", runningSolutions);
+            },
+            function (err) {
+                event.sender.send("sendingInstalledSolutions", {isError: true, message: err});
+            }
+        );
+    });
+
+    ipcMain.on("getAllSolutionsCapture", function (event, arg) {
```
amb26: Rather than manually bash on ipcMain like this in a private function, you can use the "channelUtils" that the Astea team put together as at least a half-assed kind of declarative process bridge - https://github.com/GPII/gpii-app/blob/master/src/shared/channelUtils.js#L28

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+        });
+    });
+
+    ipcMain.on("saveCapturedPreferences", function (event, arg) {
+        that.testPspChannel(arg);
+    });
+};
+
+gpii.app.captureTool.testPspChannel = function (that, pspChannel, flowManager, options) {
+    // console.log("Capturing using cloudURL: ", flowManager.settingsDataSource.options.cloudURL);
```
amb26: Remove these before commit

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,294 @@
...
+            funcName: "gpii.app.captureTool.testPspChannel",
+            args: ["{that}", "{pspChannel}", "{flowManager}", "{arguments}.0"]
+        },
+        generateDiagnostics: {
+            funcName: "gpii.app.captureTool.generateDiagnostics",
+            args: ["{that}", "{flowManager}"]
+        }
+    },
+    listeners: {
+        onCreate: [
```
amb26: Similarly, namespace and prioritise these

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
+"use strict";
+(function (fluid) {
+    var gpii = fluid.registerNamespace("gpii"),
+        ipcRenderer = require("electron").ipcRenderer;
+    fluid.registerNamespace("gpii.psp");
+    fluid.registerNamespace("gpii.captureTool");
+
+    var remoteFluid = require("electron").remote.getGlobal("fluid");
+
+    // TODO Ultimately these should be relocated to a JSON5 file, probably in
```
amb26: Mention that such a JSON5 tool would constitute what Gregg is currently referring to as the "Presentation Registry"

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+
+        },
+        modelListeners: {
+            "whatToCapture": [{
+                funcName: "console.log",
+                args: ["What To Capture: ", "{that}.model.whatToCapture"]
+            }
+            ],
+            "solutionsToCapture": [{
+                funcName: "console.log",
```
amb26: Axe these logs

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+                {
+                    func: "{that}.updateCapturedSettingsToRender",
+                    args: ["{change}.value"]
+                },
+                {
+                    func: "{that}.render",
+                    args: ["{that}.model.currentPage"]
+                },
+                {
+                    // Currently, it's a bit dicey as to replaying the binder
```
amb26: It certainly is, with no namespaces on anything!

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+        var orderedInstalledSolutions = [];
+        fluid.each(data, function (val, key) {
+            var next = fluid.copy(val);
+            next.id = key;
+            orderedInstalledSolutions.push(next);
+        });
+        fluid.stableSort(orderedInstalledSolutions, function (a, b) {
+            var idx = gpii.captureTool.appOrdering;
+            function idxOrEnd(i) {
+                return idx.indexOf(i) === -1 ? idx.length + 1 : idx.indexOf(i);
```
amb26: Why do we send them to the end?
Also, sending them to idx.length would be sufficient.
Also, avoid the double computation

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+            }
+
+            var idxA = idxOrEnd(a.id);
+            var idxB = idxOrEnd(b.id);
+            if (idxA < idxB) {
```
amb26: Accepted solution here is
return idxA - idxB;

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+        ipcRenderer.on("sendingRunningSolutions", function (event, arg) {
+            that.fullChange("runningSolutions", arg);
+        });
+
+        ipcRenderer.on("sendingAllSolutionsCapture", function (event, arg) {
+            var finalSettings = gpii.captureTool.annotateSettingsCapture(that, arg);
+
+            // Remove any solutions with zero settings per UX Review
+            fluid.each(finalSettings, function (item, key) {
+                // Is it safe to modify an object during iteration with fluid.each?
```
amb26: Use fluid.remove_if instead

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+        }
+
+        // Clear any previous capture
+        that.applier.change("capturedSettings", {}, "DELETE");
+
+        ipcRenderer.send("getAllSolutionsCapture", options);
+    };
+
+    gpii.captureTool.nextButton = function (that, currentPage) {
+        if (currentPage === "1_sign_in") { // Currently not used with UX redesign
```
amb26: Pretty regrettable hardcoded logic, but I don't see a great reason to improve it. We're not about to implement...... Spring Web Flow

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+            fluid.each(solutionData.settings, function (item, idx) {
+                fluid.remove_if(curSettings, function (curSetting) {
+                    return curSetting === solutionId + ":" + idx;
+                });
+            });
+        }
+        that.applier.change("settingsToKeep", curSettings);
+    };
+
+    gpii.captureTool.onHideShowNumSettingLink = function (vhat, event) {
```
amb26: vhat deliberate typo?

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1014 @@
...
+                checkboxEl.checked = true;
+            }
+            else {
+                checkboxEl.indeterminate = true;
+                checkboxEl.checked = false;
+            }
+        });
+    };
+
+    gpii.captureTool.onSolutionHeaderToKeepCheck = function (that, event) {
```
amb26: I'd refactor these to all just accept the target node as 2nd argument to be clearer that we don't have a dependency on the entire DOM event

src/renderer/captureTool/js/common-event-bindings.js
```diff
@@ -0,0 +1,86 @@
...
+ *         // type: jQuery <- Defaults to jQuery but could be configured ITF
+ *         method: "click",
+ *         args: ["{that}.selectProduct"]
+ *     }
+ * }
+ * ```
+ */
+fluid.defaults("gpii.binder.bindMarkupEvents", {
+    mergePolicy: {
+        decorators: "noexpand"
```
amb26: I don't see you are using this facility anywhere in your code

src/renderer/captureTool/js/capture.js
sgithens: If they aren't in the "presentation registry" list of ordered solutions, I put them on the end at the bottom of the page.

src/renderer/captureTool/js/common-event-bindings.js
sgithens: `gpii.captureTool` uses this grade

src/renderer/captureTool/js/capture.js
sgithens: done

src/renderer/captureTool/js/capture.js
sgithens: hahahahaha

src/renderer/captureTool/js/capture.js
sgithens: refactored

src/renderer/captureTool/js/capture.js
sgithens: axed

src/renderer/captureTool/js/capture.js
sgithens: commented on

src/main/dialogs/captureToolDialog.js
sgithens: done

src/main/dialogs/captureToolDialog.js
sgithens: done

src/main/dialogs/captureToolDialog.js
sgithens: added more detail to this message

src/main/dialogs/captureToolDialog.js
sgithens: Added a comment about how this is the return value from the electron promise implementation of their dialogs.

src/main/dialogs/captureToolDialog.js
sgithens: fixed

src/main/dialogs/captureToolDialog.js
sgithens: created and injected a component for this

src/main/dialogs/captureToolDialog.js
sgithens: explained, and renamed the component grade

src/main/dialogs/captureToolDialog.js
sgithens: fixed up

src/renderer/captureTool/js/common-event-bindings.js
amb26: But it doesn't apply any decorators

src/main/dialogs/captureToolDialog.js
sgithens: Removed all bashing on of icpMain and icpRenderer

src/renderer/captureTool/js/capture.js
sgithens: I don't know... I bet it would be much easier these days to implement spring web flow with our sequencing stuff :p

src/renderer/captureTool/js/common-event-bindings.js
sgithens: Oh right, renamed this since I changed the original property name from decorators to markUpEventBindings

sgithens:
@amb26 I think I've addressed the above comments, if you wanted to take another pass through this

package.json
```diff
@@ -9,12 +9,17 @@
     },
     "license": "BSD-3-Clause",
     "dependencies": {
+        "bootstrap": "3.3.7",
         "electron": "3.0.2",
         "electron-edge-js": "8.3.8",
         "electron-localshortcut": "3.1.0",
+        "gpii-handlebars": "1.1.4",
         "infusion": "3.0.0-dev.20190328T144119Z.ec44dbfab",
+        "markdown-it": "^8.4.2",
```
amb26: What is this dependency for? I see you include it in captureTool/index.html but I don't see it is used

package.json
```diff
@@ -9,12 +9,17 @@
...
         "nan": "2.10.0",
         "node-jqunit": "1.1.8",
+        "ref": "1.3.4",
```
amb26: These two dependencies seem unnecessary, they are included by windows and used there, but there shouldn't there be a need to directly reference them from gpii-app

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1121 @@
...
+     * }
+     * ```
+     * If a default value cannot be detected the return payload will be:
+     * ```json
+     * {
+     *     "hasDefault": false
+     * }
+     * ```
+     */
+    gpii.captureTool.findDefaultValue = function (schema) {
```
amb26: I guess this is the code which duplicates what is in your universal pull. To start with, all the feedback given there is good for it here, (e.g. use of fluid.get) but further we seemed to determine in our meeting today with @the-t-in-rtf that it is probably not necessary

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1121 @@
...
+
+    gpii.captureTool.updateCurrentNumberOfTotalSettingsToRender = function (applierFunc, capturedSettingsToRender) {
+        var tally = gpii.captureTool.calculateCurrentNumberOfTotalSettingsToRender(capturedSettingsToRender);
+        applierFunc("currentNumberOfTotalSettingsToRender", tally);
+    };
+
+    gpii.captureTool.updateCapturedSettingsToRender = function (that, showDefaultSettings) {
+        var togo = {};
+        fluid.each(that.model.capturedSettings, function (appData, appId) {
+            if (appData.numberOfSettings >= 0) { // sgithens DEMO_TOGGLE
```
amb26: What is DEMO_TOGGLE ?

package.json
sgithens: Right, this was actually something I was tracking down when you were giving me tips about reloading the electron window.
gpii-handlebars includes a markdown component in a grade that gets subgraded (subclassed? subinfused?) by a number of things, and so will blow up if it's not available. I actually tried overriding it in my main component once, but it shows up in other renderer infrastructure things that get created as well.

https://github.com/GPII/gpii-handlebars/blob/master/src/js/client/renderer.js#L32-L34

amb26: Please file this issue with @the-t-in-rtf

the-t-in-rtf: I just confirmed that the required markdown plugin is a direct (non-dev) dependency of gpii-handlebars, but I do recall having to do something similar in other work that relied on gpii-handlebars. 

I also just realised that we never actually cut the full release of gpii-handlebars@2.1.0, I just took care of that.  I'd strongly suggest upgrading to 2.1.0 when you get the chance, as the 2.1.x line is where I'll fix the markdown dependency issue.

package.json
sgithens: Removed!

src/renderer/captureTool/js/capture.js
sgithens: Right... I mean, in theory it's not necessary, but in practice as long as the SPI settings are structured differently these user facing apps are still going to need to make these checks somewhere to format and check the default.  
Until the day when we try to no longer keep supported settings to nice single depth values, we won't need to the sort of recursive defaults checker we talked about on the meeting. But we will need a few special handlers for SPI settings, and maybe those RGB values.  I could simplify this method and remove some of the comments to make it smaller.

I did update it to use fluid.get

src/renderer/captureTool/js/capture.js
sgithens: Oh, this was an old comment from early the UI development where we weren't sure if settings with zero settings should show up at all in the UI so I toggling this if statement. 

I've removed this comment.

sgithens:
@amb26 When you have a minute, I believe this is ready for another round of review.

package.json
amb26: Please do the gpii-handlebars update if you can - also, if this dependency still remains, please pin to an exact version

src/main/menu.js
```diff
@@ -94,6 +103,10 @@ fluid.defaults("gpii.app.menuInAppDev", {
                 func: "gpii.app.menu.generateMenuTemplate",
                 args: [
                     "{that}.model.showQSS",
+                    // "{that}.options.showCaptureTool",
```
amb26, line 106: Why did these options get commented out?

src/renderer/captureTool/index.html
```diff
@@ -0,0 +1,96 @@
...
+
+        <script type="text/javascript" src="../../../node_modules/gpii-handlebars/src/js/common/lib/delete-and-add.js"></script>
+        <script type="text/javascript" src="../../../node_modules/gpii-handlebars/src/js/common/lib/i18n.js"></script>
+        <script type="text/javascript" src="../../../node_modules/gpii-handlebars/src/js/common/equals.js"></script>
+        <script type="text/javascript" src="../../../node_modules/gpii-handlebars/src/js/common/helper.js"></script>
+        <script type="text/javascript" src="../../../node_modules/gpii-handlebars/src/js/common/jsonify.js"></script>
+        <script type="text/javascript" src="../../../node_modules/gpii-handlebars/src/js/common/md-common.js"></script>
+        <script type="text/javascript" src="../../../node_modules/gpii-handlebars/src/js/common/messageHelper.js"></script>
+        <!-- This is pretty wild, looking in to the root cause. Currently documented here as FLUID-6452: -->
+        <!-- https://issues.fluidproject.org/browse/FLUID-6452                                           -->
```
amb26, line 42: Thanks for the clear report

src/main/menu.js
sgithens: I went back and forth between whether or not to include the capture tool on the task tray menu for both the regular and dev configurations or only for `npm run dev`. At the moment I'm including them for both since we are building a custom installer for the case study.  However, once you've reviewed everything else and this is ready for master I should  remove them from the production task tray, in which case the dev menu uses these `options` rather than inheriting them from the standard menu.

package.json
```diff
@@ -35,7 +38,7 @@
         "shelljs": "0.8.2"
     },
     "optionalDependencies": {
-        "gpii-windows": "0.3.0-dev.20200317T211057Z.74cd1fa5"
+        "gpii-windows": "sgithens/windows#GPII-1908"
```
amb26, line 41: Can this be knocked back to the latest gpii-windows release of 0.3.0-dev.20200427T102955Z.2e04c69 since I don't see any pull matching this? Also, a good opportunity to merge up with master

package.json
amb26: bump

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,476 @@
...
+ * @param {gpii.flowManager.local} flowManager - The local flow manager.
+ */
+gpii.app.captureTool.channelGetInstalledSolutions = function (channelNotifier, flowManager) {
+    flowManager.capture.getInstalledSolutionsForCurrentDevice().then(
+        function (data) {
+            channelNotifier.events.sendingInstalledSolutions.fire(data);
+            var runningSolutions = {};
+            fluid.each(data, function (solutionEntry, solutionId) {
+                try {
+                    fluid.log("ABOUT TO CHECK PROCESS FOR: ", solutionId);
```
amb26, line 342: If you mean to keep these log lines in, they should have a bit more of a clue that they originated from the capture tool

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,476 @@
...
+ * @param {Array} solutionsList - Array of standard solution ID's.
+ * @return {fluid.promise} A promise resolving to the full settings capture of all the solutions
+ * in `solutionsList`.
+ */
+gpii.app.captureTool.safelyGetSolutionsCapture = function (flowManager, solutionsList) {
+    var promTogo = fluid.promise();
+    // Due to the unknown stability of some settings handlers, we are invoking the system capture
+    // on each solution individually, so if one fails the capture will still proceed.
+    var capturePromises = [];
+    fluid.each(solutionsList, function (solutionId) {
```
amb26, line 377: This is actually fluid.transform, not an each with a side-effect ....

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,476 @@
...
+            channelNotifier.events.sendingInstalledSolutions.fire({isError: true, message: err});
+        }
+    );
+};
+
+/**
+ * This performs the task of capturing solutions using the flow manager, one solution at a time.
+ * While the flow manager can handle capturing multiple solutions in one go, because there is a
+ * likelihood of unknown situations in the real world with solutions and settings handlers, we go
+ * one at a time such that if there is a failure or exception thrown in one solution, it only
```
amb26, line 364: Note that as written, you are actually issuing all of the capture operations simultaneously and not sequencing them - the comment should reflect this.

src/main/menu.js
amb26: bump

src/renderer/captureTool/js/capture.js
```diff
@@ -0,0 +1,1073 @@
...
+        "com.freedomscientific.magic",
+        "org.nvda-project",
+        "com.freedomscientific.jaws",
+        "eu.singularlogic.pixelsense.sociable",
+        "net.gpii.uioPlus",
+        "com.microsoft.windows.screenResolution",
+        "net.opendirective.maavis",
+        "com.microsoft.windows.volumeControl",
+        "net.gpii.test.speechControl",
+        "net.gpii.explode"
```
amb26, line 55: This one is not a solution that should be exposed to or mentioned to users

src/renderer/captureTool/js/common-event-bindings.js
```diff
@@ -0,0 +1,86 @@
+"use strict";
+
+/**
+ * GPII Binder Markup Events Grade
```
amb26, line 4: Add some notes here to explain i) where this code was cribbed from, ii) where it is currently in progress to be incorporated into gpii-binder. In practice this was only ever scratch-quality code and if it ever did get upgraded it would need to be given more meaningful names than "fluid.expandCompoundArg"

src/main/dialogs/captureToolDialog.js
```diff
@@ -0,0 +1,476 @@
...
+ * affects that solutions. It causes no real discernable notice in speed.
+ *
+ * @param {gpii.flowManager.local} flowManager - The local flow manager.
+ * @param {Array} solutionsList - Array of standard solution ID's.
```
amb26, line 368: Argument type String[]

package.json
sgithens: There are a few PR's from universal still needed. At the very bare minimum, it requires:
https://github.com/GPII/universal/pull/859

And then realistically for the case studies it also needs the following. One is a very small solutions fix, one is JAWS, and one is Zoomtext.
https://github.com/GPII/universal/pull/857
https://github.com/GPII/universal/pull/846
https://github.com/GPII/universal/pull/735

src/main/dialogs/captureToolDialog.js
sgithens: Updated these to more informative and less dramatic (not all caps)

src/renderer/captureTool/js/common-event-bindings.js
sgithens: Added description and links to jira ticket and current github work.

src/main/dialogs/captureToolDialog.js
sgithens: made this change

src/renderer/captureTool/js/capture.js
sgithens: removed these test GPII solutions

src/main/dialogs/captureToolDialog.js
sgithens: Updated this comment to hopefully reflect what is actually happening.

src/main/dialogs/captureToolDialog.js
sgithens: Moved this to use fluid.transform.

sgithens:
@amb26 I've believe I've addressed the comments, and this should be ready for another look over.

sgithens:
ok to test

package.json
sgithens: Ok, updated to gpii-handlebars 2.1.0 and removed the markdown-it from packages.json.

src/main/menu.js
sgithens: removed these commented items.

sgithens:
@amb26 I believe I've addressed all the bumps now

sgithens:
ok to test

sgithens:
@amb26 Ok, the tests have been fixed up and are finally passing! I believe I've addressed the remaining bumps.  If you could take another look at this, I believe it should be ready to go in.

sgithens:
Tests are passing and the merge was clean, but I'm getting an error about metrics listeners on manual startup, will need to compare and try this against master.

javihernandez:
> Tests are passing and the merge was clean, but I'm getting an error about metrics listeners on manual startup, will need to compare and try this against master.

Yes, I believe this is due to the merge of the metrics additions both in universal and windows, so gpii-app should be updated with the bits from #167

amb26:
Probably harmless and typical race condition:
````
10:52:14.048:  Test case listener has not responded after 5000ms - at sequence pos 39 of 257 sequence element {
    "changeEvent": "{that}.app.qssWrapper.qssWidget.applier.modelChanged",
    "path": "isShown",
    "listener": "fluid.identity"
} of fixture QSS Widget integration tests
````
ok to test

amb26:
And once for luck: ok to test

PR #84, opened 2019-01-28
GPII-3686: Offset windows to consider windows docked at the top or left

stegru:
When Read&Write gold is docked at the top, the QSS is positioned incorrectly.

This PR fixes it (I've only tested with magnifier)

QSS position


stegru:
https://issues.gpii.net/browse/GPII-3686

PR #87, opened 2019-02-06
GPII-3600: Add QuickStrip Mouse settings button (morphic 1.1)

krisYanachkov:
**Note:** This pull request follows the Morphic 1.1 pull request and contains all of the previous made changes that we got from that branch.

* Added new system button "Adjust Mouse"
* Adds two base versions of the Toggle and the Stepper that can be used inside a standard setting (using the node settings)
* Changed the way translation, and setting's apply process works in order to work with secondary settings list
* Added integration testing to cover the new functionality

**Important:** The current mouse settings are not properly working and need to be setup properly (after we have the proper settings for them). Related to that the integration testing will be tweaked after the appropriated settings are used.


stegru:
ok to test

amb26:
Hi there @krisYanachkov  - could you please now regenerate the pull against updated master now I have merged in Astea's #77 , it will make it much easier to see what changes are new. Also, naturally fix test and linting issues. As a "pre-review" comment, could you improve the commenting at qssDialog.js to describe the type of the first argument to gpii.app.qss.computeQssButtonsWidth explaining what values are made use of in the Object - cheers!

Karadaliev:
> 
> 
> Hi there @krisYanachkov - could you please now regenerate the pull against updated master now I have merged in Astea's #77 , it will make it much easier to see what changes are new. Also, naturally fix test and linting issues. As a "pre-review" comment, could you improve the commenting at qssDialog.js to describe the type of the first argument to gpii.app.qss.computeQssButtonsWidth explaining what values are made use of in the Object - cheers!

Done, this PR is merged with master, also I've added the typedef for the options argument.

src/main/app.js
```diff
@@ -470,15 +470,26 @@ gpii.app.changeGpiiAppShortcut = function (shortcutsManager, shortcut, oldShortc
  * @param {Object} oldValue - The previous value of the altered setting.
  * @param {String} appTextZoomPath - The path of the "App / Text Zoom" setting.
  */
+
 gpii.app.onQssSettingAltered = function (settingsBroker, appZoom, setting, oldValue, appTextZoomPath) {
+
+    function getChangeSubSetting(value, oldValue) {
+        return fluid.find_if(value.settings, function (setting, key) {
+            return !fluid.model.diff(setting, oldValue.settings[key]);
```
amb26: The algorithm here, and the one in gpii.app.qssWrapper.alterSetting have become quite convoluted - at the least, there should be greater explanation of the intent in comments, but we should have a chat on Monday about whether we can arrive at a clearer implementation. The frequent use of fluid.model.diff which is a rather heavyweight and unperformant algo is a bit of an odd smell. In practice this should be unnecessary since the ChangeApplier automatically culls the parts of changes which cause no diff.
As a stylistic point since getChangeSubSetting doesn't close over anything it should be top-level utility function, if we end up preserving it.
One route to cleaning up the workflow here and below might be using a single invocation of fluid.model.diff to compute a "changeMap" structure as seen in https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/DataBinding.js#L1763

Karadaliev: Totally agree with you, and we can implement a smarter and faster solution (we will talk about it for sure), but for now I've moved the function to a better place (a bit cleared name in line with all the other secondary settings) and I'm invoking it only when a secondary `settings` is present, not based on path values. I hope its alright as a temporary solution.

Karadaliev: This rework has been decided to be done after Morphic 1.1 release, so its not relevant anymore for this PR.

amb26: Yup, that's fine, @Karadaliev - if you can just add a comment next to this implementation referring to https://issues.gpii.net/browse/GPII-3751 then we can push it up - cheers

Karadaliev: @amb26 Done, I've set a comment in the onQssSettingAltered to remind me that this should be re-worked. I will let you know when this PR should be re-review (when I manage to transform some of the values properly).

PR #89, opened 2019-02-11
GPII-2338: Windows service

stegru:
Adds the windows service to the build


javihernandez:
ok to test

stegru:
ok to test

```
09:17:06.069:  jq: ***************

09:17:06.069:  jq: All tests concluded: 449/449 total passed in 352466ms - PASS

09:17:06.069:  jq: ***************

09:17:06.142:  Saved 673965 bytes of coverage data to '\\vboxsvr/vagrant/coverage/coverage-tests-2019-08-07T16-17-06.070Z.json'.

09:17:06.150:  Failing to resume callback for request 80vvn2ky-63267 which has already concluded
```

amb26:
Also, hasn't been merged up with master in a while

src/main/app.js
```diff
@@ -438,6 +442,18 @@ fluid.defaults("gpii.app", {
     defaultTheme: "white"
 });
 
+// Indicative fix for GPII-3818
+gpii.app.updateKeyedInUserToken = function (that, userToken) {
```
amb26, line 446: Looks like this pull has somehow had #117 incorporated - this is probably fine, but should be highlighted, and we should review how this happened

provisioning/service.json5
```diff
@@ -0,0 +1,18 @@
+// Service configuration for production.
```
javihernandez, line 1: @stegru can you incorporate the service.json5 config file that we're shipping in [morphic-1.3.x](https://github.com/GPII/gpii-app/blob/674dbb2434f8be8dc17b6ef14e3839ef16b3e0bd/provisioning/service.json5)?
Also, extra points if you:
 * update GPII_CLOUD_URL to point to stg rather than prd
 * update autoUpdate.url to point to gpii-app/master/siteconfig.json5

src/main/app.js
javihernandez: Right, I guess this was incorporated because @stegru needed to properly test the auto key-in feature, for which #117 was needed.

provisioning/service.json5
stegru: done

PR #90, opened 2019-02-15
GPII-3683: Add indication for default setting in some menus

krisYanachkov:
* Added default setting marker (star icon) for qssMenuWidget - used in Contrast and Language menus.

### DIFF
--- a/src/renderer/qssWidget/js/qssMenuWidget.js
+++ b/src/renderer/qssWidget/js/qssMenuWidget.js
@@ -263,6 +263,11 @@
                 funcName: "gpii.qssWidget.menu.presenter.applyStyles",
                 args: ["{that}", "{that}.container", "{repeater}.model.styles"]
             },
+            // Call function that set an attribute which is used for styling purporses.
+            "onCreate.defaultValue": {
+                funcName: "gpii.qssWidget.menu.presenter.defaultValue",
+                args: ["{that}.model.item.key", "{menu}.model.setting.schema.default", "{that}.container"]
+            },
             onItemFocus: {
                 funcName: "gpii.qssWidget.menu.presenter.focusItem",
                 args: [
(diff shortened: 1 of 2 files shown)


Karadaliev:
We moved the default indicator settings into a new PR because the previous one was linked to the new Adjust Mouse menu which is not necessary to say the least. This change is small and not related to the other changes we have there.

Please do the code review of this PR so we can use it as a base for the Office Simplification button too.

cindyli:
ok to test

src/renderer/qssWidget/js/qssMenuWidget.js
```diff
@@ -331,4 +336,15 @@
...
     };
+    /**
+     * Adds an attribute property for the default setting value.
+     * @param {String} key - The `key` of the setting option.
+     * @param {Object} item - The default value from the settings.
+     * @param {jQuery} container - A jQuery object representing the setting option's container.
+     */
+    gpii.qssWidget.menu.presenter.defaultValue = function (key, item, container) {
+        if (key === item) {
+            container.attr("data-default", true);
```
amb26: Why a data attribute rather than a CSS class?

Karadaliev: @amb26 Hey, well, because we noticed that the "selected" state is using attribute (aria-checked I think) too. If its not OK we can change it to class, its not a problem, just let me know.

amb26: That is a functional attribute referred to the WAI-ARIA standard - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role - DOM manipulations intended to have pure styling effects should use CSS classes, unless we can find a similar standard

Karadaliev: I can see your point, its changed to class instead.

amb26: Thanks for the fix - can you also rather than hard-coding the class name in the function block, add it to the "styles" block of the component, alongside the existing "active" style? https://github.com/GPII/gpii-app/blob/664b13ca81c8db61ae8cdc69223bd9c1971a8b8b/src/renderer/qssWidget/js/qssMenuWidget.js#L324
Also while you are doing this could you also namespace the existing bare class name "active" there with the standard prefix .fl-qssMenuWidget used by this component - cheers!

Karadaliev: Sure, I will do that after today's meeting :-)

Karadaliev: Done, both of the classes are now in the styles list and have the proper prefix.

PR #93, opened 2019-02-27
GPII-3744: Updated Build.ps1 to also install CouchDB

javihernandez:
See: https://issues.gpii.net/browse/GPII-3744
This pr depends on https://github.com/GPII/windows/pull/245


amb26:
Test case hang:
````
07:08:11.161:  Test case listener has not responded after 5000ms - at sequence pos 9 of 12 sequence element {
    "changeEvent": "{that}.app.tray.menu.applier.modelChanged",
    "path": "preferenceSetsMenuItems",
    "args": [
        "{that}.app.tray.menu.model.menuTemplate"
    ],
    "listener": "gpii.tests.app.testMenuSnapsetKeyedIn"
} of fixture GPII application integration tests
````
Looks like a genuine problem

**Comment by amb26** [comment]
ok to test

**Comment by javihernandez** [comment]
@amb26 how about this?

```
07:08:06.570:  clientService: unauthorized client with oauth2ClientId - pilot-computer
07:08:06.572:  Resetting logon change model of lifecycleManager
07:08:06.574:  UserLogonRequest fails with the error: {
    "message": "server_error while executing HTTP POST on url http://localhost:8084/access_token",
    "isError": true,
    "statusCode": 401
}
```

**Comment by cindyli** [comment]
@javihernandez this is when the client credential for "pilot-compute"  is not found in the database.

**Comment by amb26** [comment]
I can reproduce the "{that}.app.tray.menu.applier.modelChanged" error locally - or @javihernandez are you saying that one error is caused by the other? Where do you look for the server logs?

**Comment by javihernandez** [comment]
I found that here: https://ci.gpii.net/job/gpii-app-test/533/console

**Comment by amb26** [comment]
Oh yes, you're dead right - I see exactly the same failure locally

**Comment by javihernandez** [comment]
> @javihernandez this is when the client credential for "pilot-compute" is not found in the database.

but it's there, http://localhost:25984/_utils/#database/gpii/clientCredential-1 I can see:

```

{
  "_id": "clientCredential-1",
  "_rev": "1-253ae1b863144cef24617cfa7e0fa28e",
  "type": "clientCredential",
  "schemaVersion": "0.1",
  "clientId": "gpiiAppInstallationClient-1",
  "oauth2ClientId": "pilot-computer",
  "oauth2ClientSecret": "pilot-computer-secret",
  "revoked": false,
  "revokedReason": null,
  "timestampCreated": "2017-11-21T18:11:22.101Z",
  "timestampRevoked": null
}
```

Am I wrong? Maybe using the wrong config when running tests?

**Comment by cindyli** [comment]
I encountered the same "clientService: unauthorized client with oauth2ClientId - pilot-computer" error when working with universal. It was because execution of integration tests and manually starting GPII using [manualTesting config files](https://github.com/GPII/universal/tree/master/gpii/configs) are sharing the same CouchDB docker container. 

The execution of integration tests reprovision database in between each run and leave CouchDB with the test data for the last test in the end. When manually starting GPII, the same CouchDB will be re-used instead of being re-provisioned. At this point, CouchDB only has the test data rather than a full load of initial GPII data. This is why client credential data is missing from CouchDB. This problem was solved by removing "gpii" database or removing couchdb data container after running tests and before manually starting GPII.

I'm not familiar with the gpii-app test sequences. But this issue might be caused by the similar out-of-sync database in between tests.

**Comment by the-t-in-rtf** [comment]
> Am I wrong? Maybe using the wrong config when running tests?

So, two questions:

1. Which config are you using?
2. What do you mean by "running tests"?  Manual testing, or acceptance tests?  If it's the latter, which tests?

File: tests/configs/gpii.tests.all.config.json5 [hunk 1]
```diff
@@ -43,6 +43,6 @@
         }
     },
     "mergeConfigs": [
-        "%gpii-universal/gpii/configs/gpii.config.untrusted.development.all.local.json5"
+        "%gpii-universal/gpii/configs/gpii.config.untrusted.development.manualTesting.json5"
```
amb26: Binding to "manualTesting" is what is causing the current test failure since the db will not be correctly reprovisioned within runs. However, this exposes wider problems in the config binding across the whole project - see top-level comment.

amb26:
In looking over this with @the-t-in-rtf it seems clear that at least one root cause of this problem is that we have bound against the wrong config in this pull's change to "tests" - at https://github.com/GPII/gpii-app/pull/93/files#diff-41b520a9ae7e86b829f2b38c0a971f92R46 we have bound to "%gpii-universal/gpii/configs/gpii.config.untrusted.development.manualTesting.json5" which is the wrong config. However, this indeed the corresponding config to the abolished "all.local" config so this exposes that the set of configs that gpii-app is binding to needs to be revised and cleaned up.
Another awkward problem, e.g. is that the core config "app.base" binds to a production core config - https://github.com/GPII/gpii-app/blob/13d7e292a763fb3fcd10317e661aeb9f2802e940/configs/app.base.json#L25 - we do indeed want the feature described in the impl comment for https://github.com/GPII/universal/blob/master/gpii/configs/gpii.config.untrusted.production.dynamicDR.json5 that "resets the system on system startup" but calling this a "production" feature isn't too accurate.

Note that the core FlowManager "production" config currently has these two effects which we probably don't want to couple together - https://github.com/GPII/universal/blob/master/gpii/node_modules/flowManager/configs/gpii.flowManager.config.production.base.json5
All users of the gpii app, whether in "production" or "dev", will most likely want the "reset at start" functionality since this is what the app is specified to do, but they not all want the HTTP endpoints suppressed since I know that Gregg, kavya and others do a lot of manual testing with these endpoints.

javihernandez:
Changes in gpii-app configs:

* app.dev: This one runs the gpii-app dev config (which includes the dev menu) and the entire architecture locally.
* app.testing: This one doesn't include the dev menu, runs the entire architecture locally and resets to STD at startup.
* app.production: This runs the system in production mode, this is:
** It runs the untrusted flowManager locally
** It uses the cloudBased flowManager running in the cloud
** It suppresses the http endpoints and resets to STD at startup

Note that both app.dev and app.testing configs will initialize couchDB in case the DB doesn't exist.

@amb26, @cindyli, @the-t-in-rtf, 

do these last updates make sense to you? Anything else you think that we should consider doing at this point?

javihernandez:
> CI job failed: https://ci.gpii.net/job/gpii-app-tests/626/

```
"cmd": "docker ps -a --filter label=gpii-couchdb-test-harness --format '{{ json . }}'"
```

Hmm, looks like it's not picking up the env variable that way.

javihernandez:
ok to test

amb26:
getUserUsbDrives failure seen twice now - JIRAd at https://issues.gpii.net/browse/GPII-3807

amb26:
ok to test

the-t-in-rtf:
> Note that both app.dev and app.testing configs will initialize couchDB in case the DB doesn't exist.

Just to confirm, by that you mean you have the cleanDbs option set for the gpii-couchdb-test-harness?  I want to confirm that you didn't need to change the CouchDB launch process.

configs/app.testing.json
```diff
@@ -1,15 +1,11 @@
...
-                "record": "gpii.app.dev.gpiiConnector",
-                "target": "{that gpiiConnector}.options.gradeNames"
-            }
-        }
+        "gradeNames": ["fluid.component"]
     },
     "mergeConfigs": [
-        "%gpii-app/configs/app.base.json"
+        "%gpii-app/configs/app.base.json",
+        "%flowManager/configs/gpii.flowManager.config.resetAtStart.base.json5",
```
the-t-in-rtf: If we're only talking about DB content, this should not be needed.  I assume this is to reset the OS state?

the-t-in-rtf: Just for my own understanding, I tracked it back through to the [flow manager](https://github.com/GPII/universal/blob/9f07ae63cd0a8b2027453dff5f33b70a45065e5a/gpii/node_modules/flowManager/src/FlowManager.js#L74) to confirm that this is [resetting the OS to the default settings](https://github.com/GPII/universal/blob/9f07ae63cd0a8b2027453dff5f33b70a45065e5a/gpii/node_modules/flowManager/src/FlowManager.js#L299).

javihernandez:
> Just to confirm, @javihernandez, by that you mean you have the cleanDbs option set for the gpii-couchdb-test-harness?

No, I don't want the db to be re-populated every time I run the code (if that's what you are asking about and what the cleanDbs options does). My understanding is that right now, if there's no db, it should be created at startup, right?
At least, this is the behavior I expect when performing manual testing.

configs/app.testing.json
javihernandez: Right, this config makes the system to trigger a "Reset to STD" at gpii-app startup.
You can read a bit more about this functionality [here](https://github.com/GPII/universal/blob/master/documentation/ResetComputer.md).

the-t-in-rtf:
Hi, @javihernandez.  When the gpii-couchdb-test-harness encounters a CouchDB instance that lacks a given database, it always creates and then provisions it, that's correct.   [By default in the underlying gpii-couchdb-test-harness component](https://github.com/the-t-in-rtf/gpii-couchdb-test-harness/blob/cc7f87dd2eef875cbfe4cdc3c2086c36d407f9ac/src/js/harness.js#L333), databases are cleaned every time.  However, the "manualTesting" grades set the key option to false, so as long as you're using one of those, you're fine.

javihernandez:
@the-t-in-rtf thanks for the clarification. Yes, I guess we're all fine, let's get this one merged and we'll figure out later if we need anything else on this area.

javihernandez:
ok to test

javihernandez:
ok to test

PR #94, opened 2019-02-28
GPII-3738: Quick Folders

Karadaliev:
* created new widget called quickFolders to handle the new button
* added the settings, messageBundles and etc. in order to show the button and to work properly
* added new variable in the siteconfig under qss.morphicQuickFolderPath which is used as a base folder


Karadaliev:
@amb26 This one is ready to be reviewed, thanks again for the help with the siteConfig ;-).

src/renderer/qssWidget/js/qssQuickFoldersWidget.js
```diff
@@ -0,0 +1,173 @@
...
+/* global fluid */
+
+"use strict";
+(function (fluid) {
+    var gpii = fluid.registerNamespace("gpii");
+
+    /**
+     * QSS Quick folders widget
+     */
+    fluid.defaults("gpii.qssWidget.quickFolders", {
```
amb26: Widget needs some tests!
Consult https://github.com/GPII/gpii-app/blob/master/tests/QssTestDefs.js for pattern

src/renderer/qssWidget/js/qssQuickFoldersWidget.js
```diff
@@ -0,0 +1,173 @@
...
+                labelTitle: "{quickFolders}.model.messages.label",
+                searchButtonLabel: "{quickFolders}.model.messages.searchButtonLabel",
+                alertLabel: "{quickFolders}.model.messages.alertLabel"
+            },
+            errorMessageSelector: "{that}.options.selectors.errorMessage",
+        },
+
+        listeners: {
+            onCreate: {
+                func: "{that}.onCreate"
```
amb26: No need for a single-purpose invoker here - better is to just write out the invocation inline as a namespaced listener - 
````
listeners: {
    "onCreate.hideErrorMessage": {
        func: "{gpii.qssWidget.quickFolders}.errorMessage.onSuccess.fire"
    }
}

File: src/renderer/qssWidget/js/qssQuickFoldersWidget.js [hunk 3]
```diff
@@ -0,0 +1,173 @@
...
+        components: {
+            searchField: {
+                type: "gpii.psp.widgets.textfield",
+                container: "{that}.dom.searchField",
+                options: {
+                    model: {
+                        value: "{gpii.qssWidget.quickFolders}.model.value"
+                    },
+                    invokers: {
+                        onSearch: {
```
amb26: This should be an event, as the "onXxxx" naming suggests, and this impl added as a listener, perhaps with namespace "impl"

src/renderer/qssWidget/js/qssQuickFoldersWidget.js
```diff
@@ -0,0 +1,173 @@
...
+            },
+            searchButton: {
+                type: "gpii.psp.widgets.button",
+                container: "{that}.dom.searchButton",
+                options: {
+                    model: {
+                        label: "{gpii.qssWidget.quickFolders}.model.messages.searchButtonLabel"
+                    },
+                    invokers: {
+                        onClick: "{gpii.qssWidget.quickFolders}.searchField.onSearch"
```
amb26: Again onClick should be an event

src/renderer/qssWidget/js/qssQuickFoldersWidget.js
```diff
@@ -0,0 +1,173 @@
...
+                container: "{that}.dom.errorMessage",
+                options: {
+                    model: {
+                        label: "{gpii.qssWidget.quickFolders}.model.messages.alertLabel"
+                    },
+                    styles: {
+                        alertHidden: "fl-qssSearchWidget-alert-hidden"
+                    },
+                    invokers: {
+                        onError: {
```
amb26: onError and onSuccess should be events, with these bodies added as listeners. Namespaces for these impls can be "show" and "hide", the implementation is small enough that can be expressed inline, e.g.
listeners: {
    "onError.hide": {
        "this": "{that}.container",
        method: "removeClass",
        args: "{that}.options.styles.alertHidden"
    }

src/renderer/qssWidget/js/qssQuickFoldersWidget.js
Karadaliev: The onClick invoker seems to be required in the gpii.psp.widgets.button itself, as you can see in - widgets.js:320.

Karadaliev:
All of the code changes are up, working on the tests now. I will write when the tests are ready and its ready for a re-review.

src/renderer/qssWidget/js/qssQuickFoldersWidget.js
amb26: In that case we should convert the button's invoker to an event too - invoker named "onClick" with an implementation of fluid.identity is a classic anti-pattern : P
We then convert the button's "onCreate.addClickHandler" to invoke "{that}.onClick.fire"

Karadaliev: @amb26 I totally agree that its a broken pattern, that's for sure! And I've tried your method, which seems great - it seems like the button will continue working as is, and we can use it as a event in the current widget. But its not like that, I can make it work somehow in the current widget, but all other buttons are stopping working. Of course that can be fixed, by reworking all of the broken buttons, but I'm not sure if this cannot be moved into another task and just close and merge this PR as is? We can discuss it in today's meeting, I will put it into the agenda.

amb26:
ok to test

src/renderer/qssWidget/js/qssQuickFoldersWidget.js
amb26: Written up as https://issues.gpii.net/browse/GPII-3776 for the next cycle

Karadaliev:
@amb26 can you do a quick re-review here, because seems everything is ready to be merged.

amb26:
Got some conflicts now - please remerge, sorry to miss this on the previous cycle

PR #95, opened 2019-03-06
GPII-3772: Change the zoom button's tooltip

Karadaliev:
A simple change of the tooltip message of the App/Text Zoom button


amb26:
ok to test

PR #96, opened 2019-03-06
GPII-3521: QuickStrip is not fully visible on the screen

krisYanachkov:
Added missing margin to the close button width, which solves the low-resolution issue.


amb26:
ok to test

PR #103, opened 2019-03-14
GPII-3794: Changing the screen capture button's commands

Karadaliev:
* Changed the screen capture behavior from full screen, to selected region


Karadaliev:
@amb26 Only the commands has been changes, so it should be ready to merge as soon as possible 😈

PR #110, opened 2019-03-19
GPII-3796: The change language warning keeps appearing all the time

krisYanachkov:
- Show notification warning only one time per session

- Notification is not showing when user is keyed in.

### DIFF
--- a/src/main/qss.js
+++ b/src/main/qss.js
@@ -99,6 +99,7 @@ fluid.defaults("gpii.app.qssWrapper", {
 
         isKeyedIn: false,
         keyedInUserToken: null,
+        notificationShown: false,
         settings: "{that}.options.loadedSettings",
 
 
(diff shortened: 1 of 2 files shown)


javihernandez:
ok to test

amb26:
Massive stream of app zoom failures - 
````
08:19:45.940:  jq: FAIL: Module "gpii.tests.windows.appZoom" Test name "Send zoom" - Message: WM_KEYUP should only be received if testing keys
08:19:45.940:  jq: Source:     at pok (\\vboxsvr\vagrant\node_modules\infusion\tests\test-core\jqUnit\js\jqUnit.js:112:15)
    at Object.assertTrue (\\vboxsvr\vagrant\node_modules\infusion\tests\test-core\jqUnit\js\jqUnit.js:145:13)
````
I think I'll comment these out upstream until @stegru can take a look at https://issues.gpii.net/browse/GPII-3802

**Comment by amb26** [comment]
ok to test

File: src/main/qss.js [hunk 1]
```diff
@@ -353,16 +354,21 @@ fluid.defaults("gpii.app.qssWrapper", {
  * @param {Object} updatedSetting - The `gpii.app.qssNotification` instance.
  */
 gpii.app.qssWrapper.showRestartWarningNotification = function (that, qss, qssNotification, updatedSetting) {
-    if (updatedSetting.restartWarning && !that.model.disableRestartWarning) {
-        var description = fluid.stringTemplate(that.model.messages.restartWarningNotification, {
-            settingTitle: updatedSetting.schema.title
-        });
+    // The notification is shown only once, and only if its not a keyIn even
```
javihernandez: typo - s/even/event

tests/QssTestDefs.js
```diff
@@ -237,46 +237,48 @@ var navigationSequence = [
...
-            "Restart warning notification is not shown when disabled by user setting",
-            "{that}.app.qssWrapper.qssNotification.model.isShown"
-        ]
-    },
-
-    { // bring everything back to normal
-        func: "{that}.app.resetAllToStandard"
-    }
-];
+// This test are commented because of changes in GPII-3796 request.
```
javihernandez: Does this test need to be reworked according to the new implemented behavior?

javihernandez:
Hey @krisYanachkov, thanks for updating this. However, if "disableRestartWarning" is set to false, then the restart warning should be shown once at least, right? Can we also get this functionality tested or is there anything that I'm missing that makes this hard to be tested?

krisYanachkov:
> Hey @krisYanachkov, thanks for updating this. However, if "disableRestartWarning" is set to false, then the restart warning should be shown once at least, right? Can we also get this functionality tested or is there anything that I'm missing that makes this hard to be tested?

Hi @javihernandez, thanks for the question! The only thing that makes testing this functionality difficult is the lack of control over the sequence of tests. Because of this reason, the message is shown at least once before executing this test. Most likely, if I spend more time I will think of something, but in this case, I decided to focus on the new functionalities.

javihernandez:
Thanks for your clarification @krisYanachkov, can you add a single comment to the test sequence that points to this JIRA ticket I just created? https://issues.gpii.net/browse/GPII-3943

PR #112, opened 2019-03-21
GPII-3813: Customize the list of the buttons on the QSS

Karadaliev:
Adding the option to pick which button to show in QSS:
* Created `id` attribute to all option buttons in settings.json
* Added new buttonList array in siteConfig > qss node (all buttons listed)
* Added isButtonList() and filterButtonList() in main/utils.js for filtering the settings
* Used the new filtering functions in loadSettings()
* Added a few checks here and there to be sure we have the proper values before doing anything; Most of them just for missing settings (now when we can have missing ones) like the language one


cindyli:
ok to test

Karadaliev:
After the last merge (the Open USB PR) new conflicts are present, I will take care of them tomorrow and will be ready for re-test and re-review.

Karadaliev:
Done, merged properly, added the new configuration for the Open USB as well. @amb26 ready for code review.

src/main/common/utils.js
```diff
@@ -188,6 +188,78 @@ gpii.app.isPointInRect = function (point, rectangle) {
            rectangle.y <= point.y && point.y <= rectangle.y + rectangle.height;
 };
 
+/**
+ * Checks if the buttonList attribute exists in the siteConfig object
+ * @param {Object} siteConfig - instance of the siteConfig object
+ * @returns {Boolean}
+ */
+gpii.app.isButtonList = function(siteConfig) {
```
amb26: Better name "hasButtonList"

src/main/common/utils.js
```diff
@@ -188,6 +188,78 @@ gpii.app.isPointInRect = function (point, rectangle) {
...
+
+/**
+ * Looks for a `id` and matches it to the provided string
+ * return empty array when there is no button found
+ * @param {String} buttonId - the `id` of the button
+ * @param {Array} availableButtons - the full settings list
+ * @returns {Array}
+ */
+gpii.app.findButtonId = function(buttonId, availableButtons) {
+    var returnData = [];  // returning empty array by default
```
amb26: Why an empty array when you return a single button if it is found?

src/main/common/utils.js
```diff
@@ -188,6 +188,78 @@ gpii.app.isPointInRect = function (point, rectangle) {
...
+                          // siteConfig, added in the same order
+        afterList = [],   // all the buttons that don't have `id` at all,
+                          // they are added at the end of the list
+        tabindex = 100;   // starting tabindex, adding +10 of each new item
+
+    // creating the matchedList
+    // looking for `id` and if matches adding it
+    fluid.each(siteConfigButtonList, function(buttonId) {
+        var matchedButton = gpii.app.findButtonId(buttonId, availableButtons);
+        if (matchedButton !== []) {
```
amb26: I feel sure that if test cases were written for this functionality, one would find that this comparison does not work! Since [] !== []

src/main/common/utils.js
Karadaliev: Done

src/main/common/utils.js
```diff
@@ -188,6 +188,78 @@ gpii.app.isPointInRect = function (point, rectangle) {
...
+ * return empty array when there is no button found
+ * @param {String} buttonId - the `id` of the button
+ * @param {Array} availableButtons - the full settings list
+ * @returns {Array}
+ */
+gpii.app.findButtonId = function(buttonId, availableButtons) {
+    var returnData = [];  // returning empty array by default
+
+    if (fluid.isValue(buttonId) && fluid.isValue(availableButtons)) {
+        fluid.each(availableButtons, function(button) {
```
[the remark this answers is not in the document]
Karadaliev: Done, and the function is renamed a bit more clear.

src/main/common/utils.js
Karadaliev: Yeah, no sense to return empty array, returning false now.

src/main/common/utils.js
Karadaliev: True, checking if its not false now (based on the change above).

Karadaliev:
Code review changes are up, ready for re-review.

src/main/common/utils.js
```diff
@@ -190,6 +190,78 @@ gpii.app.isPointInRect = function (point, rectangle) {
...
+ * return empty array when there is no button found
+ * @param {String} buttonId - the `id` of the button
+ * @param {Object[]} availableButtons - the full settings list
+ * @returns {Boolean|Object} - returns false when button is not found,
+ * or the setting object if found
+ */
+gpii.app.findButtonById = function(buttonId, availableButtons) {
+    if (fluid.isValue(buttonId) && fluid.isValue(availableButtons)) {
+        return fluid.find_if(availableButtons, function(button) {
+            if (fluid.isValue(button.id)) {
```
amb26: This guard can be removed since you already checked that the buttonId is a value

Karadaliev: The check above (line 211) is for the param buttonId, the one here (line 213) checks if the current button in the foreach has `id` before it compares it to the buttonId itself.

The `id` node its not required, and a few buttons don't have one (all of the service buttons for example), so the check on line 214 throws errors when encounters one.

amb26: It could never throw an error, since the JS === operator just works by reference equality. Try removing the if statement and see what happens :)
Also, please run the branch through the linter - cheers

Karadaliev: OK, OK, let's say that I agree 😸 . All of the changes are up, including the lint ones.

amb26:
Failed during creation of VM - ok to test

amb26:
Looks like the chocolatey package registry is down - https://chocolatey.org/api/v2/ - no point retrying until it returns

amb26:
Looks like it's just timing out rather than down, let's have another go - ok to test

PR #114, opened 2019-03-25
GPII-3820: Change the screen capture button's tooltip

Karadaliev:
* Screen Capture button's tooltip changes


javihernandez:
ok to test

PR #115, opened 2019-03-25
GPII-3819: Change reset to standard logic

krisYanachkov:
- After reset to standard, use the value from the siteConfig to set the default language in QuickStrip.


amb26:
ok to test

src/main/gpiiConnector.js
```diff
@@ -770,7 +770,7 @@ fluid.defaults("gpii.app.dev.gpiiConnector.qss", {
         "http://registry\\.gpii\\.net/common/highContrastTheme": { value: "regular-contrast" },
         "http://registry\\.gpii\\.net/common/selfVoicing/enabled": { value: false },
         // use the initial value of the language as default setting
-        "http://registry\\.gpii\\.net/common/language": { value: "{systemLanguageListener}.model.configuredLanguage" }
+        "http://registry\\.gpii\\.net/common/language": { value: "{that}.options.defaultPreferences.systemDefaultLanguage" }
```
amb26: Please add a comment here to list the JIRA number, and add a comment to indicate that this is a provisional fix related to understanding why the default value isn't taken from the core's logic (as per meeting discussion on 25/3/19) and that we need to circle back at some point to remove this special case

amb26:
It's not clear why this is really a fix for G1 issue for GPII-3819 - since this adjusts the idea of the default rather than the current value - unless there is some special processing on the app's side that occurs on a reset to standard

amb26:
@cindyli confirmed that this pull works because the standard payload on a reset to standard is empty - in which case the actual displayed value is taken from the defaults.

siteconfig.json5
```diff
@@ -3,6 +3,9 @@
     // `true` if the button should be hidden and `false` otherwise.
     hideQssSaveButton: false,
 
+    // File containing default values which are used after resetting user settings
+    defaultValues: "C:\\Users\\vagrant\\AppData\\Roaming\\gpii\\defaultSettings.json5",
```
javihernandez: I'd say we need to use the %appdata% env variable to come up with the real path to the file.

Karadaliev: Done :-)

Karadaliev:
@javihernandez can you re-run the test to see if will pass now?

src/main/common/utils.js
```diff
@@ -211,3 +211,14 @@ gpii.app.openUSB = function() {
...
+
+/**
+ * Uses environment's %appdata% variable and combines it with the data from the site config
+ * the result should be something like:
+ * C:\Users\vagrant\AppData\Roaming\gpii\defaultSettings.json5
+ * @param {String} defaultSettingsPath
+ * @returns {String}
+ */
+gpii.app.compileAppDataPath = function(defaultSettingsPath) {
+    return process.env.appdata + "\\" + defaultSettingsPath;
```
javihernandez: better to use [path.join](https://nodejs.org/api/path.html#path_path_join_paths)

siteconfig.json5
```diff
@@ -4,7 +4,7 @@
     hideQssSaveButton: false,
 
     // File containing default values which are used after resetting user settings
-    defaultValues: "C:\\Users\\vagrant\\AppData\\Roaming\\gpii\\defaultSettings.json5",
+    defaultValues: "gpii\\defaultSettings.json5",
```
javihernandez: based on how you named this, I would think that it refers to the values itself rather than a filepath or a filename.

javihernandez: oh sorry, I see that you're using this name in the siteconfig file, then I'd say it's okay.

javihernandez:
ok to test

src/main/common/utils.js
Karadaliev: Done!

siteconfig.json5
Karadaliev: I can see your confusion, so I just added a couple of comments above, to make it a bit clearer.

Karadaliev:
@javihernandez seems like this one can't run the automated tests at this time because it requires the default settings file (the one specified in siteConfig) in order to get the standard settings. And with this file not present in the test environment it fails. Let me know if you have thoughts about how to work around that... because we do need to have this one merged before the Volume and Brightness ones (they work with default values as well).

javihernandez:
@Karadaliev I see.
Not sure what is the best approach here, but here I go.

Can't we provide our own defaultSettings.json5 file within gpii-app's testData folder? Then, we can: 

1) Make use of distribute options to point the flowManager to the defaultSettings.json5 file that we want to use. The flowManager reads this value from here https://github.com/GPII/universal/blob/master/gpii/node_modules/flowManager/src/DefaultSettingsLoader.js#L24
2) Tell gpii-app to use the gpii-app's defaultSettings.json5 file, either through distribute options or by hardcoding the value in the test

However, I'm not sure if the universal config we use for running the gpii-app tests is ready for this use case. Also, not sure if resetting to standard when running the tests is going to affect the results of the tests and/or break anything.

@amb26, @cindyli what are your thoughts on this?

tests/fixtures/siteconfig.json5
```diff
@@ -12,5 +12,6 @@
     	resizable: true,
     	alwaysOnTop: true
     },
-    surveyTriggersUrl: "http://localhost:8083/surveys/triggers.json"
+    surveyTriggersUrl: "http://localhost:8083/surveys/triggers.json",
+    defaultValues: "gpii\\defaultSettings.json5"
```
cindyli: This name `defaultValues` is confusing. It might be more reasonable using a name like "defaultSettingsFileLocation".

cindyli:
@Karadaliev, a couple of ideas after looking at this pull request:
1. Create your own defaultSettings.json5 for running tests somewhere in /tests directory and point [tests/fixtures/siteConfig.json5 to read it](https://github.com/GPII/gpii-app/blob/bc80b1bf6d7e677124a1744b8bb4965134d8d68f/tests/fixtures/siteconfig.json5#L16);
2. A more complex way is to use the approach I used in universal: split the tests of loading defaultSettings.json5 and applying these settings separately. This will require [the value of `loadedSettings`](https://github.com/GPII/gpii-app/blob/bc80b1bf6d7e677124a1744b8bb4965134d8d68f/src/main/gpiiConnector.js#L808) to be configurable.

In case you choose (2), the approach I use for testing default settings feature in the universal is to test the loading of defaultSettings.json5 file and the application of settings read from defaultSettings.json5 separately.

The latter in the universal is relatively easier. There is [this unit test for the default settings loader](https://github.com/GPII/universal/blob/master/gpii/node_modules/flowManager/test/DefaultSettingsLoaderTests.js).

The former is more complex. Using [the reset at startup in the untrusted environment](https://github.com/GPII/universal/blob/master/tests/UntrustedResetAtStartTests.js) as an example, the test takes these steps:
1. Create [an own config file](https://github.com/GPII/universal/blob/master/tests/UntrustedResetAtStartTests.js#L21) to [override the read of defaultSettings.json5 from the file system](https://github.com/GPII/universal/blob/master/tests/configs/shared/gpii.tests.resetDefaultSettings.config.json5#L4-L15);
2. [Pass desired settings](https://github.com/GPII/universal/blob/master/tests/shared/ResetDefaultSettingsTestDefs.js#L79-L87) you'd like to test into each test definition.

javihernandez:
FYI, https://github.com/GPII/universal/pull/803 has been merged, which means that now the PSP should report the default schema settings and that we no longer need to make gpii-app to look for the file and read the default settings itself. Can you rework this PR to fit this new situation?

Since gpii-app and windows are out-of-sync at this moment for some reasons, in order for you to keep going with this PR you can make your package.json file to point to this version of windows ("javihernandez/windows#cf9d38fb2baef3b62ad73f43c7a075712f5c4898") until the two repos are synced again. Sorry for any inconvenience.

amb26:
With the merge of @cindyli 's work for GPII-3828 in the core under https://github.com/GPII/universal/pull/820 I believe that this pull can now be abandoned - if noone disagrees, I will close it

Karadaliev:
> 
> 
> With the merge of @cindyli 's work for GPII-3828 in the core under [GPII/universal#820](https://github.com/GPII/universal/pull/820) I believe that this pull can now be abandoned - if noone disagrees, I will close it

We talked about with @javihernandez today and indeed you can close this PR.

PR #117, opened 2019-03-25
GPII-3818: "Indicative fix" for auto-keyin race condition

amb26:


javihernandez:
> CI job failed: https://ci.gpii.net/job/gpii-app-tests/1079/

Lint step failed, the closing bracket of the `distributeOptions` block was commented during your last update with current master.

PR #120, opened 2019-03-26
GPII-3822: QuickStrip not re-scaling on the fly

Karadaliev:
* changing the scale factor only if its called properly from the QSS wrapper


javihernandez:
ok to test

PR #121, opened 2019-03-29
GPII-3837: Remove Light and Dark Gray from High Contrast menu

krisYanachkov:
- Remove Light and Dark Gray from High Contrast menu


cindyli:
ok to test

PR #123, opened 2019-04-02
GPII-3840: Screen Zoom Menu dots clipped on the larger resolutions

Karadaliev:
* Very small UX adjustments to the stepper and menu widgets


javihernandez:
ok to test

PR #124, opened 2019-04-17
GPII-3872: Add siteConfig option for QSS open on startup

Karadaliev:
Added option to show QSS on startup:
* Added new siteConfig value called showQssOnStart (boolean)
* Using set value to determine to show QSS or not on Morphic's startup


javihernandez:
ok to test

Karadaliev:
@amb26 @javihernandez Can you take a look at this one as well?

PR #127, opened 2019-05-09
GPII-3882: New button - launch DocuMorph.exe

Karadaliev:
* Added new type of service button for the executing a file
* Added new service button for the Launch DocuMorph to QSS


Karadaliev:
@javihernandez Can you send me the proper path to the DocuMorph executable when you know where it will be installed? Because right now for testing purposes I'm executing the path to Notepad ;-).

javihernandez:
ok to test

Karadaliev:
@javihernandez can you run the tests once more?

javihernandez:
ok to test

javihernandez:
@Karadaliev can you please resolve conflicts with master?

src/main/common/utils.js
```diff
@@ -196,10 +197,31 @@ gpii.app.isPointInRect = function (point, rectangle) {
...
+
+/**
+ * A custom function for handling opening of an .exe file.
+ * @param {String} executablePath - path to executable file
+ */
+gpii.app.launchExecutable = function (executablePath) {
+    var fileProperties = fs.statSync(executablePath);
+
+    // Check that the file is executable
+    if (fileProperties.mode === 33206) {
```
amb26: Please use octal literal here so it is easier to understand what the required file mode is

src/renderer/qss/js/qssServiceButtons.js
```diff
@@ -206,6 +206,29 @@
         openUSB.fire();
     };
 
+    /**
+     * Inherits from `gpii.qss.buttonPresenter` and handles interactions with the "launch .exe file"
+     * QSS button.
+     */
+    fluid.defaults("gpii.qss.launchExePresenter", {
```
amb26: This default is named generically, but it is hard-wired to only launch the docuMorphExecutable option from the site config. Either find a way to express what it launches parametrically (e.g. by having the particular instance of it supplying an option "docuMorphExecutable" or else the resolved siteConfig option), or rename it

Karadaliev:
Code review changes are applied. Everything is greatly simplified, and the file mode is represented in octal literal.

src/main/common/utils.js
Karadaliev: Done

src/renderer/qss/js/qssServiceButtons.js
Karadaliev: The presenter is renamed properly, and now its using a universal function to execute the file.

PR #131, opened 2019-06-03
GPII-3944: Provide alt-text for the "X" that closes the QSS

Karadaliev:
* Added aria-label to the QSS's close button
* Added arial-label to the widget's close button


javihernandez:
ok to test

PR #132, opened 2019-06-03
GPII-3838: Change Open USB Drive per new specifications (v2)

Karadaliev:
* Removed the old service button, and all of his events
* Created new widget to handle the new spec, with the buttons inside - one to mount USB, and one to eject them all
* Re-used the old mount function, and added new one for eject and attached to the widget's events


javihernandez:
ok to test

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+
+    /**
+     * Represents the QSS USB widget.
+     */
+    fluid.defaults("gpii.qssWidget.openUSB", {
+        gradeNames: ["fluid.viewComponent", "gpii.psp.heightObservable", "gpii.psp.selectorsTextRenderer"],
+        model: {
+            disabled: false,
+            setting: {},
+            messageChannel: "LoadInitialOfficeRibbonsState", // Channel listening for messages related to usb mount, unmount functionality
```
amb26: Why are we listening to an OfficeRibbonsState channel when this is the USB widget?

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+            menuControls: ".flc-qssOpenUSBWidget-controls",
+            footerTip: ".flc-qssOpenUSBWidget-footerTip"
+        },
+        enableRichText: true,
+        activationParams: {},
+        closeDelay: 1200,
+        components: {
+            repeater: {
+                // whenever the setting changes (e.g. if the change is made via the PSP)
+                type: "gpii.psp.repeater",
```
amb26: Why is there a repeater here when there is a static design with 2 buttons?

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+    var gpii = fluid.registerNamespace("gpii");
+
+    /**
+     * Represents the QSS USB widget.
+     */
+    fluid.defaults("gpii.qssWidget.openUSB", {
+        gradeNames: ["fluid.viewComponent", "gpii.psp.heightObservable", "gpii.psp.selectorsTextRenderer"],
+        model: {
+            disabled: false,
+            setting: {},
```
amb26: Document the expected contents of the model here

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+
+    /**
+     * Creates an array of possible values for the given setting to be provided to the
+     * `gpii.psp.repeater` instance.
+     * @param {Object} setting - the current setting for the QSS menu.
+     * @return {Object[]} - An array of key-value pairs describing the key and the value
+     * (the visible name) of the setting.
+     */
+    gpii.qssWidget.openUSB.getRepeaterItems = function (setting) {
+        var schema = setting.schema || {},
```
amb26: There doesn't seem to be a setting or a schema associated with the openUSB functionality - what is this function used for?

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+
+    /**
+     * Gets the value the pressed button from the menu and determines which even to fire
+     * in case of no valid button type is pressed does nothing
+     *
+     * @param {String} value - accepts only "Mount" and "Unmount"
+     * @param {Component} openUSB - The `gpii.qssWidget.openUSB` instance
+     * @param {KeyboardEvent} keyboardEvent - The keyboard event (if any) that led to the
+     * change in the setting's value.
+     * @param {EventListener} mountUsbEvent - handle to the onQssOpenUsbRequested event
```
amb26: Not an eventListener but an event. Write {fluid.event} for type of these

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+     * @param {EventListener} unmountUsbEvent - handle to the onQssUnmountUsbRequested event
+     * @param {String} messageChannel - The channel to which the message should be sent.
+     */
+    gpii.qssWidget.openUSB.handleOpenUSB = function (value, openUSB, keyboardEvent, mountUsbEvent, unmountUsbEvent, messageChannel) {
+        if (fluid.isValue(value)) {
```
amb26: No particular need for a switch for a 2-element choice - also no point in failing silently since this represents a design error.
How about
if (value === "Mount") {
 ....
} else if (value === "Unmount") {
  ....
} else {
    fluid.fail("Unrecognised command to handleOpenUSB: " + value)
}

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+        },
+        events: {
+            onItemFocus: "{repeater}.events.onItemFocus"
+        },
+        listeners: {
+            "onCreate.applyStyles": {
+                funcName: "gpii.qssWidget.openUSB.presenter.applyStyles",
+                args: ["{that}", "{that}.container", "{repeater}.model.styles"]
+            },
+            onItemFocus: {
```
amb26: I notice that this listener is now repeated identically in 3 places - qssMenuWidget.js, qssScreenCaptureWidget.js and now here - time to factor this and all the things it depends on into a shared grade

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+    });
+
+    /**
+     * Focuses the current QSS menu option if its index matches the specified `index` parameter.
+     * @param {Component} that - The `gpii.qssWidget.menu.presenter` instance.
+     * @param {focusManager} focusManager - The `gpii.qss.focusManager` instance. for the QSS.
+     * @param {jQuery} container - A jQuery object representing the setting option's container.
+     * @param {Number} index - The index of the setting option to be focused.
+     */
+    gpii.qssWidget.openUSB.presenter.focusItem = function (that, focusManager, container, index) {
```
amb26: More cut and paste from the other 2 locations - rewrite so this just appears once

src/renderer/qssWidget/js/qssOpenUsbWidget.js
```diff
@@ -0,0 +1,350 @@
...
+                    listeners: {
+                        "onCreate.enable": {
+                            this: "{that}.container",
+                            method: "removeClass",
+                            args: ["{that}.options.styles.disabled"]
+                        }
+                    }
+                }
+            },
+            closeTimer: {
```
amb26: This block also part of the cut and paste from qssMenuWidget. Refactor as with the focus material referenced below

src/renderer/qssWidget/js/qssOpenUsbWidget.js
krisYanachkov: My mistake.

Karadaliev:
@amb26 ready for re-review.

amb26:
@krisYanachkov actually the CI failed the test for a genuine reason this time:
````
07:17:56.560:  RendererCoverage: Report collected: 1gpii.app.pspInApp
07:17:56.560:  RendererCoverage: Still awating: [
    "gpii.app.qssMorePanel"
]
````
followed by a test hang waiting for renderer coverage of the last component. I have seen this kind of thing locally - we should review with @the-t-in-rtf if we can understand the race condition that can be responsible.
I merged the pull anyway since it passed locally and the cause of the failure is unrelated to the pull.

PR #138, opened 2019-07-01
GPII-3705: Reduced the chances of the tray button not showing.

stegru:
Creating the window for the tray button sometimes failed. This fix makes it keep retying to created it.

Somehow, the taskbar gets into a state where child windows can't be added to it - it's in this state when the start menu is shown, and could also be like this at some point during the login.

To test: Open the start menu while starting Morphic - the icon doesn't appear. Close the start menu, and it will appear.


trayButton/tray-button.c
```diff
@@ -832,21 +832,33 @@ int CALLBACK WinMain(HINSTANCE hInst, HINSTANCE hPrevInst, LPSTR cmd, int show)
...
+				// Also, creating without a parent but using SetParent returns ERROR_INVALID_PARAMETER, under the same
+				// conditions.
+				DWORD thisError = GetLastError();
+				if (thisError != lastError) {
+					fail("CreateWindowEx (retrying)");
+					lastError = thisError;
+				}
+				Sleep(1000);
+			}
+			// Continue trying to create the window if it didn't succeed.
```
javihernandez: Maybe stupid question, but, what if we never succeed in creating the tray button? 
Bear in mind that my comment goes beyond saying "let's add a timeout" and raise an error if we can't create the tray icon after X retries. Is it acceptable for gpii-app to keep starting up even if we can't create the tray icon? 
Of course, this is just an observation, not blocking the merge of this PR at all, but maybe something that we should think about it in case we find out that even with this patch, we can't guarantee 100% that the tray icon is going to be there. :man_shrugging:

stegru: I think there should be a timeout, and then show the normal tray icon after that.

javihernandez: Good point! In any case, my vote goes for shipping this patch and see how effective it is.

stegru:
ok to test

stegru:
ok to test

```
default: [NuGet] Could not connect to the feed specified at 'https://chocolatey.org/api/v2/'.
Please verify that the package source (located in the Package Manager Settings) is valid
and ensure your network connectivity.
```

stegru:
ok to test

PR #144, opened 2019-07-12
GPII-3967: Create new My Saved Settings button

krisYanachkov:
- Added the new saved settings widget


krisYanachkov:
I will add tests for the widget as soon as possible.

src/main/app.js
```diff
@@ -430,6 +431,10 @@ fluid.defaults("gpii.app", {
             funcName: "gpii.app.resetAllToStandard",
             args: ["{that}", "{psp}", "{qssWrapper}.qss"]
         },
+        reApplyPreferences: {
+            funcName: "gpii.app.reApplyPreferences",
```
amb26: No need for this whole definition and function , you can just bind to the LM's definition directly as follows:
reApplyPreferences: {
    func: "{lifecycleManager}.replayEnvironmentalLogin"
}

src/main/qss.js
```diff
@@ -255,7 +256,8 @@ fluid.defaults("gpii.app.qssWrapper", {
             type: "gpii.app.qssWidget",
             options: {
                 model: {
-                    scaleFactor: "{qssWrapper}.model.scaleFactor"
+                    scaleFactor: "{qssWrapper}.model.scaleFactor",
+                    lastEnvironmentalLoginGpiiKey: "{qssWrapper}.model.lastEnvironmentalLoginGpiiKey"
```
amb26: Nice use of iterated relays

src/main/qss.js
```diff
@@ -99,6 +99,7 @@ fluid.defaults("gpii.app.qssWrapper", {
             restartWarningNotification: null
         },
 
+        lastEnvironmentalLoginGpiiKey: null,
```
amb26: I'd leave this blank, with just a comment, on the off chance that your "null" value gets propagated back into the system on startup

krisYanachkov:
@amb26  Seems that the "lastEnvironmentalLoginGpiiKey" is not filled at the moment of the creation of "qssWrapper" and the value always is undefined. I need to find a workaround for this. I'll appreciate any hints :)

amb26:
Well, we expect the value to remain undefined until the environmental logon actually occurs. This would require testing with a configuration in which that listener is activated - however given our lack of time before the installer is required it's probably going to be hard to devote time to this.

krisYanachkov:
> Well, we expect the value to remain undefined until the environmental logon actually occurs. This would require testing with a configuration in which that listener is activated - however given our lack of time before the installer is required it's probably going to be hard to devote time to this.

As a temporary solution, I can try to get the actual value when the my saved widget is open through ipc as I do [https://github.com/krisYanachkov/gpii-app/blob/9526714c960d2602d2d5ab3e3db36e32c8ff92d3/src/renderer/qssWidget/js/qssVolumeWidget.js#L58](here) . What do you think about it?

amb26:
Yes, you're correct - currently things are not set up so that the relay can occur across the process boundary, you will have to apply some IPC by hand

krisYanachkov:
@amb26 The IPC communication is done. Tested on the last build, everything seems stable.

amb26:
ok to test

amb26:
Failure: 
````
08:28:15.127:  jq: FAIL: Module "gpii.tests.production.config tests" Test name "GPII application integration tests" - Message: Unexpected failure in test case (see following log for more details): Assertion failure - check console for more details: Error in invoker record: could not resolve members func, funcName or method to a function implementation - got undefined from {

    "func": "{lifecycleManager}.replayEnvironmentalLogin",
````
Looks like this is bound to an old version of the dependent projects without the new method

PR #145, opened 2019-07-13
GPII-4019: Create Translate button & change Language naming

krisYanachkov:
- G1 - Translate menu is too small -- causes scroll when screen is zoom


javihernandez:
ok to test

javihernandez:
ok to test 

```
02:19:31.802:  Test case listener has not responded after 5000ms - at sequence pos 16 of 16 sequence element {
    "task": "gpii.tests.app.instrumentedDialog.requestCoverage",
    "resolve": "fluid.log",
    "resolveArgs": "Renderer coverage collected!"
} of fixture GPII application dev config integration tests

```

Let's see what happens when we ok to test again

amb26:
This is a race condition that perhaps we see about 1 time in 10 in the gpii-app tests, it seems like a race condition either in the instrumentation library or the test cases themselves that @the-t-in-rtf has been notified of

PR #149, opened 2019-07-26
GPII-4026: Tooltips are not properly displayed/aligned

Karadaliev:
* added new tooltip styles for the arrow on the left, and on the center
* added new event onTooltipArrowDirection that handles the change of the arrow style
* in the getTooltipPosition() function we are checking where the tooltip is, moving it to fit the screen and changing the arrow position of needed

P.S. For now the QSS cannot reach the left part of the screen (it can be at most 80% size of the screen), so there is no way the arrow should be on the left. But just in case all of the preparations of that case are left and used if needed.


stegru:
ok to test

krisYanachkov:
@stegru can you re-run the test to see if will fail again?

javihernandez:
ok to test

src/main/dialogs/quickSetStrip/qssTooltipDialog.js
```diff
@@ -57,6 +63,16 @@ fluid.defaults("gpii.app.qssTooltipDialog", {
                     "{that}.model.setting"
                 ]
             }
+        },
+        arrowDirection: {
+            target: "arrowDirection",
+            singleTransform: {
```
amb26: Very odd - why is there any transform here given that the effect is to return the supplied value? Seems like this is a noop

src/main/dialogs/quickSetStrip/qssTooltipDialog.js
```diff
@@ -157,11 +188,31 @@ gpii.app.qssTooltipDialog.showIfPossible = function (that, setting, btnCenterOff
...
+        // setting the offset to fit the screen
+        offsetX = screenWidth - tooltipWidth;
+        // changing the arrow to be in the center
+        arrowDirection = availableDirections.centerDirection;
+    } else {
+        arrowDirection = availableDirections.defaultDirection;
+    }
+
+    // apply the new arrow direction
+    that.applier.change("arrowDirection", arrowDirection);
```
amb26: This is an odd side-effect in a function whose purpose is to return the offsets structure. I'm guessing that the original intention was to implement this inside the getArrowDirectionFunction - I suggest that we go either wholesale with a model-driven approach or else continue with this getXxx approach and then instead add the arrowDirection into the returned payload from this function

amb26: I imagine it is impossible for the tooltip direction to change during the lifetime of the displayed tooltip, right?

src/main/dialogs/quickSetStrip/qssTooltipDialog.js
Karadaliev: Because we do need to go through the modelRelay, but didn't know how to make it work without the singleTransform part :-). Any guidance?

src/main/dialogs/quickSetStrip/qssTooltipDialog.js
Karadaliev: Yeah, I do agree that it's strange to have applier in a get type function. Probably will be easier to just return the value and set it elsewhere, I will take a look.

And yes, I'm pretty sure the direction won't change during one instance of the tooltip.

Karadaliev: **Update:** I've made a slight change (your second recommendation) to where the direction is set.

Karadaliev:
@amb26 maybe retest this? The tests died with "Connection refused - connect(2) for "127.0.0.1" port 5985" message... 😈

amb26:
ok to test

krisYanachkov:
@amb26 can we retest this?

amb26:
ok to test

src/main/dialogs/quickSetStrip/qssTooltipDialog.js
amb26: Still very unclear - why does this transform still exist?

src/main/dialogs/quickSetStrip/qssTooltipDialog.js
```diff
@@ -173,12 +222,15 @@ gpii.app.qssTooltipDialog.getTooltipPosition = function (that, btnCenterOffset)
...
+    var offsetAndDirection = gpii.app.qssTooltipDialog.getTooltipPosition(that, btnCenterOffset);
 
     that.dialog.setAlwaysOnTop(true);
 
     // reposition window properly
-    that.setPosition(offset.offsetX, offset.offsetY);
+    that.setPosition(offsetAndDirection.offsetX, offsetAndDirection.offsetY);
+
+    // apply the new arrow direction
+    that.applier.change("arrowDirection", offsetAndDirection.direction);
```
amb26: Given you have folded all of this into the "show" action there doesn't seem to be a need to modelise this any more

src/renderer/qssTooltipPopup/js/qssTooltipPopup.js
```diff
@@ -47,16 +54,43 @@
...
+                                "{arguments}.0",
+                                "{qssTooltipPopup}.model.availableDirections",
+                                "{qssTooltipPopup}.container"
+                            ]
                         }
                     },
                     events: {
-                        onTooltipUpdated: null
+                        onTooltipUpdated: null,
+                        onTooltipArrowDirection: null
```
amb26: Doesn't seem to be a need for this event - if you are continuing to modelise this in the popup, you can just add gpii.qss.qssTooltipPopup.onTooltipArrowDirection as a model listener

krisYanachkov: Right now we are determining the arrow direction in the main process. That's why we are having this event, which is called when the direction is changed in the main process. And in the renderer this applies the proper styles. So what do you mean by that?

Karadaliev:
@amb26 Can you take at this one again, because seems like all of the code review changes are done, and its merged with the latest master now 😈 .

Karadaliev:
@amb26 reminder that we are still waiting on new review changes, or merge on this one too 🥇

PR #151, opened 2019-08-26
GPII-4079: Custom user button for opening apps and urls

Karadaliev:
* Create couple of new button presenters for the custom buttons - "custom-launch-app" and "custom-open-url"
* Re-using the existing openCloudFolderPresenter() (now renamed to openUrl) and launchExecutable() methods to either open url, or execute a program
* Created generateCustomButton() method which uses the JSON object from the siteConfig's buttonList to replicate a button like the real ones from settings.json
* Created validation routine for the both button types. If the validation fails the button name is changed and the button itself its disabled. For the APP button it checks if the file exists and if its a executable; for the WEB button tries to do a regex validation of the url

**Temporary:**
* Added one WEB, and one APP button in siteConfig to illustrate how it works (it will be removed before merging)
* Added a Chrome full-screen option to illustrate how the maximize should work (set alwaysUseChrome to true, and in the custom button's data fullScreen to true in order to see it in action)


Karadaliev:
@amb26 Please start the review process when you have the time 🛩

amb26:
ok to test

src/main/common/utils.js
```diff
@@ -219,6 +220,72 @@ gpii.app.findButtonById = function (buttonId, availableButtons) {
...
+        serviceButtonTypeWeb = "custom-open-url",
+        titleAppNotFound = "Program not found",
+        titleUrlInvalid = "Invalid URL",
+        disabledStyle = "disabledButton",
+        data = false;
+
+    // we need to have the data, with all required fields:
+    // buttonId, buttonName, buttonType, buttonData
+    if (buttonData && fluid.isValue(buttonData.buttonId) && fluid.isValue(buttonData.buttonName) && fluid.isValue(buttonData.buttonType) && fluid.isValue(buttonData.buttonData)) {
+        var buttonType = buttonData.buttonType === "APP" ? serviceButtonTypeApp : serviceButtonTypeWeb;
```
amb26: We should supply some helpful diagnostic if not all the data is present. You might consider rolling a little utility like https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/Fluid.js#L226

src/main/common/utils.js
```diff
@@ -219,6 +220,72 @@ gpii.app.findButtonById = function (buttonId, availableButtons) {
...
+        if (fluid.isValue(buttonData.popupText)) {
+            // adding the tooltip text as well
+            data.tooltip = buttonData.popupText;
+        }
+        if (fluid.isValue(buttonData.fullScreen) && buttonData.fullScreen) {
+            // adding the full screen option if there is one
+            data.schema.fullScreen = true;
+        }
+        if (buttonData.buttonType === "APP") {
+            // checks if the file exits and its executable
```
amb26: typo exists

src/main/common/utils.js
```diff
@@ -347,3 +421,49 @@ gpii.app.getVolumeValue = function (browserWindow, messageChannel) {
...
+};
+
+/**
+ * Simple function to try to validate the url
+ * @param {String} url - browser's url
+ * @return {Boolean} - returns `true` when the url looks valid
+ */
+gpii.app.checkUrl = function (url) {
+    // validating the url
+    var pattern = new RegExp("^(https?:\\/\\/)?" + // protocol
```
amb26: Give a source from where you copied this code :)

src/renderer/common/js/utils.js
```diff
@@ -85,18 +85,60 @@
...
 
-    /*
+    /**
+     * A custom function for handling activation of the "Quick Folders" QSS button.
+     * opens a provided url in the default browser using electron's shell
+     * @param {String} siteUrl - cloud folder's url
+     * @param {Boolean} alwaysUseChrome - true to use chrome, rather than the default browser.
+     * @param {Boolean} forceFullScreen - the function requires the browser to be open maximized
+     */
+    gpii.psp.openUrl = function (siteUrl, alwaysUseChrome, forceFullScreen) {
```
amb26: Stick this in a Windows namespace to make it clear that we need to write a replacement for it once MacOS support starts to appear

Karadaliev:
@amb26 Unfortunately Steve's PR its still not merged so in order this to work we need to use his package.

src/main/common/utils.js
amb26: bump - please give some attribution

Karadaliev: I've added the source before, I wonder what happened with it -- **re-added** now.
_(also second time typing this answer...)_

amb26:
I've rolled 0.3.0-dev.20190930T092641Z.74b8070 release of gpii windows including @stegru 's GPII-4079 fix - try updating to it and see how it goes

Karadaliev:
> 
> 
> I've rolled 0.3.0-dev.20190930T092641Z.74b8070 release of gpii windows including @stegru 's GPII-4079 fix - try updating to it and see how it goes

Yup, seems to work now -- updated in package.json.

PR #152, opened 2019-08-26
GPII-3848: Remove PSP code

Karadaliev:
* Removed all of the translations related to the PSP from the message bundles
* Removed all of the PSP related files, both js, css, html, and so on
* Removed the usage of the PSP as a whole
* Altered all of the tests related to PSP (some of them removed)
* Added `id` to all of the settings.json buttons and used in the siteConfig
* Applied the button ids to the button's html code for easier testing
* Separating (in progress) of the QSS buttons into different files for each widget for cleaner testing process, and to be easy to add new buttons without affecting the old ones (no more conflicts and such). The new tests use the same button ids as well to pinpoint the button in the QSS instead of the previous indexing


Karadaliev:
@amb26 We are still moving and redesigning the QSS button tests, but you can start the review process as well, its a lot to look at 💯

amb26:
ok to test

amb26:
I think this work is looking excellent - reorganising the tests, as you say, will be a bit meaty

Karadaliev:
@amb26 I know its a huge PR, and hard to read through all of it, but I do hope to merge this one as fast as possible in order to use it as a base of all other major tasks (and tech debt, like the new Electron). So please start the CR process when you have the time! 🛩

amb26:
Great work on eliminating this dangerous mass of legacy code!

PR #153, opened 2019-09-05
GPII-3776: Replace use of onXxxx invokers in gpii-app with events + listeners

Karadaliev:
Replacing onClick invokers with events through the whole QSS.


amb26:
ok to test

src/renderer/psp/js/widgets.js
```diff
@@ -306,7 +309,10 @@
             "onCreate.addClickHandler": {
                 "this": "{that}.container",
                 method: "click",
-                args: ["{that}.onClick"]
+                args: ["{that}.events.onClick.fire"]
+            },
+            onClick: {
```
amb26: Since the listeners do not have a namespace, this definition should be removed

amb26:
This pull now needs updating

PR #154, opened 2019-09-09
GPII-4001: Naming of change source "settingAlter" in GPII app is unclear

Karadaliev:
* Changed 'settingAlter' source namespace to 'fromWidget' in every widget


amb26:
ok to test

PR #157, opened 2019-10-15
GPII-4192: Change the default font of Morphic

Karadaliev:
Changed the default font to "Segoe UI":

* Changed the default font everywhere
* Removed the OpenSans font files
* Removed fonts.css (which was used to define OpenSans) and all of his usages


amb26:
ok to test

PR #158, opened 2019-10-15
GPII-3600: Add QuickStrip Mouse settings button

Karadaliev:
**Note:** This is the most recent version of the Adjust Mouse task, we closed the old ones.

* Added new Adjust Mouse button
* Added secondary settings option in the settings.json, the translations and Undo
* Modified the alterSettings to accommodate the new secondary settings  (including [GPII-3751](https://issues.gpii.net/browse/GPII-3751))
* Modified the updateChangeIndicator in order to accommodate the new secondary settings

**Important:** We will add the tests to this PR as soon they are ready.


cindyli:
ok to test

Karadaliev:
@amb26 everything is fixed here (and fully operational), can we start the review process? 🛩

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,715 @@
...
+                funcName: "gpii.qssWidget.mouseSpeedStepper.clearElementsAnimation",
+                args: [
+                    "{that}.dom.stepperButton",
+                    "{that}.options.styles"
+                ]
+            }
+        },
+
+        invokers: {
+            activateIncBtn: {
```
amb26: We've now got a great opportunity here to remove all this code which is now duplicated in 3 places - 
i) here
ii) qssStepperWidget.js
iii) qssVolumeWidget.js

Please consolidate into a shared grade

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,715 @@
...
+                options: {
+                    model: {
+                        label: "{mouseSpeedStepper}.model.messages.decrementButton"
+                    },
+                    invokers: {
+                        activate: "{mouseSpeedStepper}.activateDecBtn"
+                    }
+                }
+            },
+            // similar to qssWidget component
```
amb26: Even this comment itself is copied from qssStepperWidget, and qssVolumeWidget!

testData/qss/settings.json
```diff
@@ -177,6 +177,65 @@
...
+                "id": "mouseSpeed",
+                "path": "http://registry\\.gpii\\.net/applications/com\\.microsoft\\.windows\\.mouseSettings.PointerSpeed",
+                "schema": {
+                    "type": "number",
+                    "default": 10,
+                    "min": 1,
+                    "max": 20,
+                    "divisibleBy": 2
+                },
+                "value": 10,
```
amb26: What is the separate meaning to the "value" entry here, in addition to the default value of the schema? I'm trying to remember why we didn't make use of the schema default value for the existing ordinary settings.

Karadaliev: Yeah... this is legacy thing, its used in all of the current settings. We will try to remove it from settings.json (I mean the generic "value" there) and use schema.default instead. I will keep you posted on this.

src/renderer/qssWidget/js/qssMouseWidget.js
Karadaliev: This is true, we will look into it, but probably will create a tech debt type task specifically for this, because its a bit outside the scope of this PR (it affects few other buttons). We will discuss it tomorrow after today's meeting.

src/renderer/qss/js/qssSettingButtons.js
```diff
@@ -49,14 +48,26 @@
      * on whether the changes to the setting can be undone and whether the new value of
      * the setting is the same as its default one.
      * @param {Component} that - The `gpii.qss.settingButtonPresenter` instance.
-     * @param {Object} setting - The setting object corresponding to this QSS button.
-     * @param {Any} value - The new value of the `setting`.
+     * @param {Object} setting - The setting object corresponding to this QSS button
```
amb26: Since this structure has now become significantly complex, please can you add a @typedef JSdoc entry for it explaining all the structures it can have, and update any references in function arguments

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,444 @@
...
+     * @param {Object} schema - Describes the schema of the setting.
+     * @param {Number} schema.min - The minimum possible value for the setting.
+     * @param {Number} schema.max - The maximum possible value for the setting.
+     * @param {Number} schema.divisibleBy - The amount which is added or subtracted
+     * from the setting's value every time this function is invoked.
+     * @param {Boolean} shouldSubtract - Whether the `divisibleBy` amount should be
+     * subtracted from or added to the setting's value.
+     * @return {Boolean} Whether there was a change in the setting's value.
+     */
+    gpii.qssWidget.mouse.makeRestrictedStep = function (that, value, schema, shouldSubtract) {
```
amb26: Let's try and unify this with baseStepper.makeRestrictedStep. What is the nature of the edge case which gives rise to restrictedValue of 2 - can we fix this by simply adjusting the schema before passing it into this function?
Other irregularities
 - "shouldSubtract" can easily be represented using "stepMultiplier" scheme from the other function
 - why do we use a precision of 2 here and 3 in the other function?

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,444 @@
...
+                args: "{that}.model.item.indicatorValue"
+            }
+        }
+    });
+
+    /**
+     * Alters the custom element attribute in order to change the styles applied to it.
+     * @param {jQuery} indicatorContainer - The container for the indicator element
+     * @param {Object} stateAttribute - Options for the state defining attribute
+     * @param {Object} indicatorData - The condition data for the element
```
amb26: Please add @typedef structures for these two argumens

amb26: Actually this whole function can be axed as identical to gpii.qssWidget.baseStepper.indicator.updateState - please check for other duplications

src/renderer/qssWidget/js/qssBaseStepperWidget.js
```diff
@@ -0,0 +1,478 @@
...
+                func: "{gpii.qssWidget.baseStepper.indicators}.events.onIndicatorClicked.fire",
+                args: "{that}.model.item.indicatorValue"
+            }
+        }
+    });
+
+    /**
+     * Alters the custom element attribute in order to change the styles applied to it.
+     * @param {jQuery} indicatorContainer - The container for the indicator element
+     * @param {Object} stateAttribute - Options for the state defining attribute
```
amb26: Please add @typedef structures for these two arguments

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,444 @@
...
+
+    /**
+     * Handler for a single indicator element.
+     *
+     * Each indicator element has three states: normal, selected and default.
+     * These three states are indicated using a custom html element
+     * attribute - "data-type". Depending on the state of this attribute, different
+     * styles are applied (refer to the CSS for more info).
+     */
+    fluid.defaults("gpii.qssWidget.mouse.indicator.presenter", {
```
amb26: This entire grade is redundant with "gpii.qssWidget.baseStepper.indicator.presenter"

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,444 @@
...
+     * Generates the different indicators' data based on a setting .
+     * indicators are generated using the setting's `min`, `max` and `divisibleBy` properties.
+     * In case either of those is missing, no indicators will be generated.
+     * Note that items will be recomputed every time the setting changes but only items that
+     * need to be re-rendered will do so (changeApplier merges the values).
+     * @param {Object} setting - The setting for which indicators must be created
+     * @return {Object[]} - The list of data for each indicator element. In case no indicators
+     * can be generated an empty array is returned
+     */
+    gpii.qssWidget.mouse.getIndicatorsList = function (setting) {
```
amb26: As well as being mostly identical with the baseStepper version, this impl contains a dangerous floating-point bounded loop. If we definitely need to handle decimal values, the only safe approach is to multiply up the divisibleBy step by 100 or 1000 etc. so that it becomes an integer, otherwise we run the risk of missing the upper loop bound due to floating point error. Whilst fixing both these impls, also try to unify them so that the special case for value 0 is factored out somehow - why is it that 0 is treated specially by the mouse settings?

src/renderer/qssWidget/js/qssVolumeWidget.js
```diff
@@ -198,326 +190,5 @@
...
-    };
-
-
-    /**
-     * The QSS stepper widget for Volume button
-     *
-     * Represents the quick set strip stepper widget. It is used for
-     * incrementing/decrementing a setting.
-    */
-    fluid.defaults("gpii.qssWidget.volumeStepper", {
```
amb26: Some good savings here due to the refactoring

amb26:
Whilst I've left some review comments here for improving consistency and readability, I'm happy to sign off this branch as suitable to roll into an installer if we find we need to roll this functionality into a release this month. There are some minor conflicts with master again.

src/renderer/qssWidget/js/qssMouseWidget.js
krisYanachkov: We not reusing the original one from the base stepper grade because the windows mouse speed setting cannot accept a value of 0 and the logic need to be slightly different. 
Also, we use a precision of 3 in the base stepper because we need that precision for the volume setting value.

amb26: But as far as I can see, the value 0 is not within the valid schema range?
https://github.com/GPII/universal/blob/master/testData/solutions/win32.json5#L8184

krisYanachkov: Yes, but because in this case, the stepper uses even multiplier. The setting min value is odd and broke the calculations.

Karadaliev:
@amb26 Looks like we are ready for re-review 🛩

src/renderer/qssWidget/js/qssMouseWidget.js
amb26: I think we could fold this into a general purpose algorithm which always guarantees that a restricted value is a multiple of the step - it would be very helpful to axe this bulky and largely duplicated function.

Karadaliev: Sooo, about that - the reason we are doing this is because of the problem with mouseSpeed not accepting 0 as a value (basically doing nothing). So if we are using the default logic (1 to 20, division by 2) then when launching the app and clicking "Faster" the values will be 1, 3, 5, and at the end 20 (because will hit the max value). When going back the value will be even 20, 18, 16 and eventually 1 (because will hit the low value). This by itself its not that horrible, but the indicators itself are created initially and they will follow the second logic (20, 18, 16...) so it will work only when hit the max value first.

I'm sure there is a possible way this to be solved by some math solution, but we don't have the time to do it right now. And with that we have 2 solutions - the first one is to leave it as it is; to change the schema from 2 to 20 (or from 1 to 19) just to make it even. So... pick your poison :smiley_cat:

amb26: Don't worry - the poison will not be that bad :)
var multiplier = 1000;
var adjustedMin = Math.ceil(multiplier * schema.min/ schema.divisibleBy) / multiplier;
var adjustedMax = Math.floor(multiplier * schema.max / schema.divisibleBy) / multiplier;
You can then harmonise your parseFloat line to this scheme by adjusting it to 
value = Math.round(multiplier * (value - step)) / multiplier;
This rounding will not be perfect but should be good enough for the current purposes, and express the intention consistently.

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,230 @@
...
+     * by clicking on it or pressing "Space" or "Enter"). What this function
+     * does is to change the `enabled` model property to its opposite value.
+     * @param {Component} that - The `gpii.psp.widgets.switch` instance.
+     * @param {Component} toggleWidget - The `gpii.qssWidget.mouseWidgetToggle` instance.
+     * @param {fluid.event} event - onQssWidgetSettingAltered event
+     */
+    gpii.qssWidget.mouseWidgetToggle.toggleModel = function (that, toggleWidget, event) {
+        if (toggleWidget.model.setting.schema.mapOff || toggleWidget.model.setting.schema.mapOn) {
+            if (that.model.enabled) {
+                toggleWidget.model.setting.value = toggleWidget.model.setting.schema.mapOff;
```
amb26: These updates need to be made via the ChangeApplier API or a model relay, rather than adjusting the model in place

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,230 @@
...
+            }
+        },
+
+        components: {
+            toggleButton: {
+                type: "gpii.psp.widgets.switch",
+                container: "{that}.dom.toggleButton",
+                options: {
+                    model: {
+                        enabled: {
```
amb26: Having an expander which reads model values on startup is liable to races depending on when the source component gets initialised, and doesn't take care of keeping the values synchronised subsequently. It looks like this application would best be handled using the valueMapper transform, of which there is a short example in the docs: https://docs.fluidproject.org/infusion/development/ModelTransformationAPI.html#example-1-shorthand-with-primitive-output-values

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,230 @@
...
+        }
+    });
+
+    /**
+     * Transforms a number value to boolean.
+     * @param {Number} value - The value of the setting.
+     * @param {Object} schema - Describes the schema of the setting.
+     * @return {Boolean} The modified value.
+     */
+    gpii.qssWidget.mouseWidgetToggle.transformValue = function (value, schema) {
```
amb26: We can expect to axe this function moving over to the valueMapper

testData/qss/settings.json
amb26: Please write this up as a separate JIRA we can track after this pull goes in

src/renderer/qssWidget/js/qssMouseWidget.js
```diff
@@ -0,0 +1,223 @@
...
+                options: {
+                    model: {},
+                    modelRelay: {
+                        "enabled": {
+                            target: "enabled",
+                            singleTransform: {
+                                type: "fluid.transforms.valueMapper",
+                                defaultInput: "{mouseWidgetToggle}.model.value",
+                                match: {
+                                    500: false,
```
amb26: Strange to map these literal values 500, 5000 to false and true, I am guessing these correspond to some values from the schema in which case they should be referenced from there directly

src/main/qss.js
```diff
@@ -606,35 +618,62 @@ gpii.app.qssWrapper.populateLanguageSettingOptions = function (settingOptions, l
 
 gpii.app.qssWrapper.loadSettings = function (assetsManager, installedLanguages, locale, messageBundles, settingOptions, settingsFixturePath, siteConfig) {
     var availableSettings = fluid.require(settingsFixturePath), // list of all available buttons
-        loadedSettings = availableSettings; // by default we are getting all of the buttons
+        loadedSettings = availableSettings, // by default we are getting all of the buttons
+        multipleir = 1000;
```
amb26: typo multipleir -> multiplier. Please add comment that this value is only supplied in order to avoid rounding errors when comparing values with the bounds

testData/qss/settings.json
amb26: Bump comment!

src/renderer/qssWidget/js/qssBaseStepperWidget.js
```diff
@@ -0,0 +1,528 @@
...
+
+
+    /**
+     * Generates the different indicators' data based on a setting .
+     * indicators are generated using the setting's `min`, `max` and `divisibleBy` properties.
+     * In case either of those is missing, no indicators will be generated.
+     * Note that items will be recomputed every time the setting changes but only items that
+     * need to be re-rendered will do so (changeApplier merges the values).
+     * @param {Object} setting - The setting for which indicators must be created
+     * @return {Object[]} - The list of data for each indicator element. In case no indicators
```
amb26: Please supply an @typedef for this structure so it is clear what data actually gets returned

src/renderer/qssWidget/js/qssBaseStepperWidget.js
```diff
@@ -0,0 +1,528 @@
...
+    });
+
+
+    /**
+     * Generates the different indicators' data based on a setting .
+     * indicators are generated using the setting's `min`, `max` and `divisibleBy` properties.
+     * In case either of those is missing, no indicators will be generated.
+     * Note that items will be recomputed every time the setting changes but only items that
+     * need to be re-rendered will do so (changeApplier merges the values).
+     * @param {Object} setting - The setting for which indicators must be created
```
amb26: I think we already have a @typedef somewhere for this argument?

src/main/qss.js
amb26: Please also add a comment explaining that this multiplier should correspond with the precision chosen in qssBaseStepperWidget

testData/qss/settings.json
Karadaliev: https://issues.gpii.net/browse/GPII-4252

Karadaliev:
@amb26 All of the code review changes requested are done, and pushed, you can take a look once more! 🛩

PR #160, opened 2019-10-16
GPII-4188: New Snipping Tool

Karadaliev:
GPII-4188: Added new service button that executes the internal Windows 10 snipping tool

**Important:** This works ONLY on Windows 10 _build 1903 or higher_.


stegru:
ok to test

src/renderer/common/js/utils.js
```diff
@@ -151,6 +151,21 @@
         return false;
     };
 
+    /**
+     * A custom function for handling executing the Snipping Tool command.
+     * @param {String} command - path to executable file
+     * @return {Boolean} - returns `true` on successfully executed command
+     */
+    gpii.windows.openSnippingTool = function (command) {
```
amb26: Can't we just use gpii.windows.launchExecutable for this?

Karadaliev: I'm not sure the reason, but this "command" don't work if its escaped by quotes (which we need for the longer paths like "Programs files" for example). And of course launchExecutable checks if the file exists (and have the proper rights), which we cannot use for this command.

So yeah, in reality we can use launchExecutable, but for that I need to add another argument in order for know if its executable file, or command. But that's basically the same as using different function.

amb26: Please add a comment to the function giving a short explanation of this

Karadaliev: Done, the explanation is added as comment on the function.

amb26:
The failure that we often see with a failed wait for one of a set of renderer coverage tests, note to @the-t-in-rtf 
````
05:04:58.999:  RendererCoverage: Still awating: [
    "gpii.app.waitDialog"
]
````
ok to test

the-t-in-rtf:
One problem I see is that [the function I expose to allow them to send coverage data](https://github.com/GPII/gpii-testem/blob/master/src/js/client/coverageSender.js#L64) doesn't do any kind of results checking.  One option would be to add at least log messages for [a range of XHR events](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#Events).

If the failure is at a different point in the chain, it wouldn't help, but at least it's an easy thing to nail down.  I have created [GPII-4202](https://issues.gpii.net/browse/GPII-4202) to track that small improvement.

the-t-in-rtf:
I just cut a dev release of gpii-testem with improved coverage sending error messages: `2.1.11-dev.20191021T135156Z.56c646b.GPII-4156`

amb26:
@krisYanachkov - could you add a dependency update to that release of gpii-testem to this pull so that we can work towards getting better diagnostics? Cheers!

PR #164, opened 2019-11-08
GPII-4227: Replaced MySavedSettingsBtn onClick invoker to be a listener

javihernandez:
See https://issues.gpii.net/browse/GPII-4227


src/renderer/qssWidget/js/qssMySavedSettingsWidget.js
```diff
@@ -61,7 +61,7 @@
                     model: {
                         label: "{mySavedSettings}.model.messages.reApplyPreferencesButtonLabel"
                     },
-                    invokers: {
+                    listeners: {
                         "onClick": "{channelNotifier}.events.onQssReApplyPreferencesRequired.fire"
```
amb26, line 65: This listener should have a namespace otherwise it can't be overridden. Something like "onClick.reApplyPreferences"

javihernandez: indeed, done

PR #165, opened 2019-11-15
GPII-4139: Create SideCar extension of the Widgets

Karadaliev:
* Added new option to add "sideCar" and "sideCarWithSettings" messages into the settings message bundles
* Added new siteConfig option qss > osSettingsAvailable that determines if the sideCar content will have direct Windows/OSX settings in the code, or it will be just explanations
* Implemented the sideCar panel to be shown (if content available) on the right side of the current widget
* Changed the "Help & More" into a multi-purpose button, if there is sideCar content acts like a Hide/Show button, if not opens the Help external link
* In the sideCar panel itself we have additional "Help & More" button that uses the original learnMoreLink data (so you can go to that url no matter if you have sideCar panel or not)


javihernandez:
ok to test

src/renderer/qssWidget/js/qssSideCart.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * The grade that handles widget's sideCart content
```
amb26: I think the name as supplied in the specs is actually "SideCar", not that it is terribly important : P

siteconfig.json5
```diff
@@ -5,6 +5,9 @@
 
     // Configuration options for the QSS window
     qss: {
+        // Whether a user can open Settings pages and Control Panels of the operating system
+        osSettingsAvailable: true,
```
amb26: Is there a JIRA for the core side of this implementation?

Karadaliev: In this case this option its not linked to core, just shows different side car information based on this variable.

amb26: Where will the variable get its information from?

Karadaliev: You see the place, in the siteconfig file itself. Its a toggle do you want to show Windows Setting direct links in the side car or not. At least this is how I read the specs.

Karadaliev:
@amb26 Can we try to merge this one ASAP? Because its used in both the Mouse and Color Vision filters and no we have a "hanging" changes, that cannot be applied neither here, nor in their branches. It can help me out greatly if this one is merged before them! 🛩

src/renderer/qssWidget/js/qssSideCart.js
Karadaliev: Yup, you are right, all of the instances of "sideCart" including variable names, comments, and file names are renamed.

src/main/dialogs/quickSetStrip/qssWidgetDialog.js
```diff
@@ -269,18 +270,22 @@ gpii.app.qssWidget.getWidgetPosition = function (that, btnCenterOffset) {
 gpii.app.qssWidget.show = function (that, setting, elementMetrics, activationParams) {
     activationParams = activationParams || {};
 
+    var scaleFactor = that.model.scaleFactor,
+        // we assume that the widget have a sideCar by default
+        hasSideCar = (fluid.isValue(setting.schema.sideCar) && setting.schema.sideCar === false) ? false : true,
```
amb26: Looks like this can simply be hasSideCar = setting.schema.sideCar !== false

src/main/qss.js
```diff
@@ -772,11 +772,26 @@ gpii.app.qssWrapper.applySettingTranslation = function (qssSettingMessages, sett
...
 
+        // sideCar
+        if (fluid.isValue(message.sideCar)) {
+            translatedSetting.sideCar = message.sideCar;
+        } else {
+            translatedSetting.sideCar = "";
+        }
+
+        // sideCar (with osSettingsAvailable set to true)
+        if (fluid.isValue(message.sideCarWithSettings) && message.sideCarWithSettings !== "") {
```
amb26: translatedSetting.sideCarWithSettings = message.sideCarWithSettings || "";

src/renderer/qssWidget/js/qssSideCar.js
```diff
@@ -0,0 +1,59 @@
...
+                args: ["{that}"]
+            }
+        }
+    });
+
+    /**
+     * Adds the appropriate data in the sideCar based on the data available and the siteconfig's
+     * osSettingsAvailable variable. If there is data in sideCarWithSettings and osSettingsAvailable
+     * is true we are showing that, if not the base sideCar message is displayed.
+     * @param  {Component} that - The instance of the widget
```
amb26: These days we try to say which kind of component - e.g. @param {gpii.qssWidget.sideCar} that

PR #168, opened 2019-12-05
GPII-4253: Made gpii-windows package an optional (OS-whitelisted) dependency

christopher-rtf:
Notes for posterity:

npm does not currently have robust support for OS-specific package dependencies in an app's package.json file.

To make the Morphic QuickStrip (gpii-app) cross-platform, the OS-specific packages have been whitelisted for their specific OS in their package.json files ("win32" for Windows, "darwin" for macOS, etc.).

These OS-specific packages have been moved from the "dependencies" section in our app's package.json file to the "optionalDependencies" section.  When executing an npm install, npm will therefore not pull in the OS-specific optionalDependencies which do not apply to the host OS.

Please note that marking our OS-specific dependencies as optional dependencies has a side-effect: if their npm installation fails, it may fail silently.  If this causes significant confusion or other issues for project development, we may want to consider a "node-gyp" approach to pulling in OS-specific dependencies.  We may also want to consider using a postinstall script which dynamically npm installs the appropriate dependency; however such a postinstall script is generally not recommended since it would self-modify the package.json file (therefore creating an unusual exception for our git check-in workflow).

### DIFF
--- a/main.js
+++ b/main.js
@@ -58,8 +58,9 @@ app.on("second-instance", function (event, commandLine) {
     }
 });
 
-// this module is loaded relatively slow
+// this module is loaded relatively slowly
 // it also loads gpii-universal
+// NOTE: if the OS-specific support package was not loaded successfully, the require(".../index.js") function will throw an exception
 require("gpii-windows/index.js");
 
 require("./index.js");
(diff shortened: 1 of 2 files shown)


amb26:
ok to test

amb26:
I think these changes are fine - even if for some reason the npm installation fails silently, we will get a pretty indicative failure on startup which will then prompt the installer to try for a verbose install. We could even in future add some specific exception handling to the require site in order to prompt this explicitly.

PR #174, opened 2020-01-08
GPII-3925: Knowing the tray button position

stegru:
For @Karadaliev, this PR is just to run it through CI, but you can just pick/merge this.

The function can be called like: `{tray}.getIconBounds()`, returns `{x:, y:, width:, height:}`


stegru:
closing... it's now somewhere in https://github.com/GPII/gpii-app/pull/177

PR #177, opened 2020-01-22
GPII-3925: Create Morphic promotion window

Karadaliev:
Added the option to create a Promotional window loaded on Morphic's start. All the options are under siteConfig> promotionWindow. And in order to see it you need to provide a url to image (png/jpg/svg) or a web page that can be loaded into the container.

Example web page that was created just for testing:
promoContentUrl: "http://www.goldenorbs.com/promo/promo-window.html"


javihernandez:
ok to test

src/main/dialogs/promotionWindowDialog.js
```diff
@@ -0,0 +1,159 @@
...
+            displaySize = { width: screen.getPrimaryDisplay().workAreaSize.width, height: screen.getPrimaryDisplay().workAreaSize.height - taskBarHeight},
+            offset = { x: displaySize.width - ((trayPosition.x + (trayPosition.width / 2)) / scaleFactor), y: displaySize.height - ((trayPosition.y - (trayPosition.height / 2)) / scaleFactor) };
+
+        // setting the position next to the tray icon
+        that.setPosition(offset.x, offset.y);
+    } else if (that.options.siteConfig.centered) {
+        // centering the promo window
+        var centeredPosition = gpii.browserWindow.computeCentralWindowPosition(that.options.siteConfig.width, that.options.siteConfig.height);
+        that.setPosition(centeredPosition.x, centeredPosition.y);
+    } else if (offsetX || offsetY) {
```
amb26: sloppy - what if they are both numerically zero? Better to check whether they are both not undefined, or else, given this is the fallback branch, simply unconditionally do that.setPosition(offsetX || 0, offsetY || 0)

Karadaliev:
@amb26 Please don't merge this one until we decide how many of the missing features are we doing this round (hopefully will be decided today).

Karadaliev:
@amb26 All of the features are added, so this PR is ready for a review! 🛩

src/main/dialogs/promotionWindowDialog.js
```diff
@@ -0,0 +1,222 @@
...
+    repositionDelay: 1900,
+
+    config: {
+        attrs: {
+            width: "{that}.options.siteConfig.width",
+            height: "{that}.options.siteConfig.height",
+            resizable: "{that}.options.siteConfig.resizable",
+            movable: "{that}.options.siteConfig.resizable.movable",
+            skipTaskbar: "{that}.options.siteConfig.resizable.skipTaskbar",
+            frame: "{that}.options.siteConfig.resizable.frame",
```
amb26: These three appear to be at the wrong path - as seen in siteconfig they are at top-level paths such as {that}.options.siteConfig.frame

src/main/dialogs/promotionWindowDialog.js
```diff
@@ -0,0 +1,222 @@
...
+        },
+        fileSuffixPath: "promotionWindow/index.html"
+    },
+
+    events: {
+        delayedClose: null,
+        onRepositioningRequired: null
+    },
+    listeners: {
+        "delayedClose": {
```
amb26: No need to quote this key

src/main/dialogs/promotionWindowDialog.js
```diff
@@ -0,0 +1,222 @@
...
+ * the window will be displayed with the default position (0, 0). If the 'centered' option
+ * is specified, the dialog will be positioned in the center of the screen.
+ * @param {gpii.app.promotionWindowDialog} that - The instance of the widget.
+ * @param {gpii.app.timer} timer - An instance of `gpii.app.timer` used for showing
+ * the widget with a delay.
+ * @param {Integer} offsetX - The x offset from the right edge of the screen.
+ * @param {Integer} offsetY - The y offset from the bottom edge of the screen.
+ * @param {gpii.app.tray} tray - The `gpii.app.tray` instance.
+ */
+gpii.app.promotionWindowDialog.show = function (that, timer, offsetX, offsetY, tray) {
```
amb26: This function is almost entirely identical with repositionPromostionWindow with the exception of the timer - please consolidate

src/main/dialogs/promotionWindowDialog.js
```diff
@@ -0,0 +1,222 @@
...
+            type: "gpii.app.timer",
+            options: {
+                listeners: {
+                    onTimerFinished: {
+                        funcName: "{promotionWindowDialog}.close"
+                    }
+                }
+            }
+        },
+        showTimer: {
```
amb26: I think we can probably consolidate all of this "show" and "reposition" machinery into one

src/renderer/common/js/utils.js
```diff
@@ -169,6 +171,34 @@
         return false;
     };
 
+    /**
+     * A custom function for retrieving a file extension.
+     * @param {String} file - path to the file.
+     * @return {String} - returns the file extension.
+     */
+    gpii.windows.getFileExtension = function (file) {
```
karamfil: Is the extra function necessary? 
```gpii.windows.getFileExtension = path.extname```

File: src/main/tray.js [hunk 7]
```diff
@@ -386,5 +393,14 @@ gpii.app.trayButton.windowMessage = function (that, hwnd, msg, wParam) {
...
             break;
         }
+    } else if (msg === that.trayButtonPositionMessage) {
+        var lParamValue = lParam.address();
+        // The bounding rectangle is 4 shorts crammed in both the message parameters.
+        that.rect = {
+            x: wParam & 0xffff,
+            y: (wParam >> 16) & 0xffff,
+            width: lParamValue & 0xffff,
+            height: (lParamValue >> 16) & 0xffff
```
karamfil: What is the idea behind this code? I see a lot of magic numbers and the shifts are a bit strange. My suggestion would be to explain this in great detail and cover fully with tests if there is no better way to write it.

src/main/tray.js
```diff
@@ -386,5 +393,14 @@ gpii.app.trayButton.windowMessage = function (that, hwnd, msg, wParam) {
         default:
```
karamfil: Is the empty default necessary?

src/main/dialogs/dialogManager.js
```diff
@@ -331,13 +340,15 @@ gpii.app.dialogManager.close = function (dialogManager, selector) {
...
  * @param {Boolean} isKeyedIn - Indicates whether there is a currently keyed
  * in user.
  */
-gpii.app.dialogManager.closeDialogsOnKeyOut = function (dialogManager, isKeyedIn) {
+gpii.app.dialogManager.closeDialogsOnKeyInOut = function (dialogManager, isKeyedIn) {
     if (!isKeyedIn) {
         dialogManager.close("survey");
+    } else {
+        dialogManager.close("promotionWindow");
     }
```
karamfil: Maybe it is possible for this to be written with a ternary operator.

src/main/dialogs/promotionWindowDialog.js
```diff
@@ -0,0 +1,192 @@
...
+        // getting the tray position only when it's required (siteConfig > positionByTrayIcon)
+        var trayPosition = tray.getIconBounds(),
+            screen = require("electron").screen,
+            taskBarHeight = screen.getPrimaryDisplay().bounds.height - screen.getPrimaryDisplay().workAreaSize.height,
+            scaleFactor = screen.getPrimaryDisplay().scaleFactor,
+            displaySize = { width: screen.getPrimaryDisplay().workAreaSize.width, height: screen.getPrimaryDisplay().workAreaSize.height - taskBarHeight},
+            offset = { x: displaySize.width - ((trayPosition.x + (trayPosition.width / 2)) / scaleFactor), y: displaySize.height - ((trayPosition.y - (trayPosition.height / 2)) / scaleFactor) };
+
+        // setting the position next to the tray icon
+        that.setPosition(offset.x, offset.y);
```
karamfil: `screen` is used only with `.getPrimaryDisplay()` so instead of using `screen`, it could be better to directly create a variable `primaryDisplay`. This way it will reduce the code. You could even further create variables for `workAreaSize` like what you did with `scaleFactor` 👍 

I am personally not very fond of the way `displaySize` and `offset` are set as objects as this feels unnecessary and will allocate more memory. Writing these as variable and ordering them in the code vertically instead of as right now horizontally would make the code a lot more readable. 

Here is an example: 

```
displayWidth = workAreaSize.width,
displayHeight = workAreaSize.height - taskBarHeight,
offsetX = displayWidth - ((trayPosition.x + (trayPosition.width / 2)) / scaleFactor),
offsetY = displayHeight - ((trayPosition.y - (trayPosition.height / 2)) / scaleFactor)
```

stegru:
ok to test

src/main/dialogs/promotionWindowDialog.js
amb26: Bump

krisYanachkov: Show and reposition were fixed a long time ago [here](https://github.com/GPII/gpii-app/pull/177/commits/1a8d1e2d8a79cf259fdb3790f31639dfbd015a08) . Let me know if there is anything else that I need to change.

PR #179, opened 2020-01-31
GPII-4065: Add resetToStandardProfileUrl to siteconfig

javihernandez:
See https://issues.gpii.net/browse/GPII-4065 and https://issues.gpii.net/browse/GPII-4296.
Depends on https://github.com/GPII/universal/pull/834

Take into account that the references are now pointing to custom branches, just for testing. I'll update them as soon as the other pieces are in place.


src/main/siteConfigurationHandler.js
```diff
@@ -46,55 +46,60 @@ fluid.defaults("gpii.app.siteConfigurationHandler", {
                     ]
                 }
             },
-            target: "{app qssWrapper}.options.settingOptions.hiddenSettings"
+            target: "{that qssWrapper}.options.settingOptions.hiddenSettings"
```
amb26: Why did all of these get changed - do you remember?

cindyli: This was changed by me at fixing the issue Javi had with overriding `{defaultSettingsLoader}.options.defaultSettingsUrl` using the url defined in the site config file. In [this fix](https://github.com/javihernandez/gpii-app/pull/4/files), `gpii.app.siteConfigurationHandler` was [moved to the top level](https://github.com/GPII/gpii-app/pull/179/files#diff-bc658f47466ba31df21972f46d3f02d4R641) where `gpii.appWrapper` is and `app` is now a sub-component of it. So I changed target from `app` to `that`.

cindyli:
@javihernandez, I issued https://github.com/javihernandez/gpii-app/pull/5 to fix the CI failure. Note that this fix is already in https://github.com/GPII/gpii-app/pull/184.

javihernandez:
> @javihernandez, I issued [javihernandez#5](https://github.com/javihernandez/gpii-app/pull/5) to fix the CI failure. Note that this fix is already in #184.

Thanks, the fix is already included as I merged current master. 
I tested it and worked great, also, tests are passing :dancer:

amb26:
But if this fix is already in #184 which is merged, why didn't this pull auto-close?

javihernandez:
> But if this fix is already in #184 which is merged, why didn't this pull auto-close?

Because Cindy made the changes in different commits against different branches, look:

- https://github.com/javihernandez/gpii-app/pull/5/commits/9a136b93323aa2408276454c9e6f9e8ec4285287 against my branch
- https://github.com/GPII/gpii-app/commit/42b3353576492bc6d7a086ce873da8d17ccd56e7 against master

amb26:
OK thanks - so we can just close this PR?

javihernandez:
> OK thanks - so we can just close this PR?

Nope, this pull includes the changes to distribute the _resetToStandardProfileUrl_ option through the flowManager for it to use a remote resetToStd profile. So it should be a matter of reviewing and (eventually) getting it merged.

PR #182, opened 2020-02-19
GPII-4367: Testing javihernandez/windows#GPII-4367

javihernandez:
Testing if this solves the build issues reported at https://github.com/GPII/gpii-app/pull/178#issuecomment-588409356

PR #178, opened 2020-01-29
GPII-3958: Adjust gpii-app to work with the new PSPChannel API.

cindyli:
With [the change of PSPChannel API](https://github.com/GPII/universal/pull/826) being merged into the universal repo, gpii-app stops working with the universal master. 

PSPChannel now expects a slightly different incoming message while PSPChannel output messages stay unchanged. The PSPChannel API is [here](https://github.com/GPII/universal/blob/master/documentation/PSPChannel.md).

This pull request adjusts gpii-app to work with the new API. @Karadaliev or @krisYanachkov, I'm not sure if it covers all messages sent to PSPChannel (Local flow manager). Please have a look. Thanks.

Please don't merge this pull request directly since it is pointing to my own gpii/windows branch which pulls in the universal master branch.

PR #183, opened 2020-02-21
GPII-4196: Create the new More panel

Karadaliev:
**More Panel is created with all of the required basic functions as follows:**

* New service button to open the More panel;
* New styles for the QSS and the More Panel (that shows above it);
* New panel (well, the More Panel) with grid-style button layout that you can put any desired button from the buttonList there, even the custom buttons;
* From a usability standpoint the More Panel works properly even changing the DPI level, or not filling properly the morePanelList data, it will auto-fill the grid with empty elements when needs to;
* Also we handle the widget window's position based on the position of the button now, so if the widget gets too close to the right part of the screen it will still fit and the arrow will show over the appropriate button;
* For the widget's sideCar we made similar arrangements (and according to the original sideCar specs as well), we make sure that the sideCar will be visible no matter where the button is, even to the extend to show the sideCar on the LEFT of the widget window if there is no space on the right.

**Usage Information:**

* The options for the More Panel grid are in siteConfig > morePanelOptions node
* The buttons that will show in the grid are in siteConfig > morePanelList. It uses a simple array structure, one element for one row (by default 3 rows, so 3 lists)
* The link and the both button blocks are HTML code that can be changed freely using the message bundles: "gpii_qss_list_infoBlock", "gpii_qss_list_buttonBlock1", "gpii_qss_list_buttonBlock2"

**IMPORTANT:**
This is fully functional version of the More Panel witch covers GPII-4337, GPII-4274, GPII-4313, and parts of GPII-4337 (the too close to the right scenario). It's great base for testing and the review process and can be used for demo purposes. Still **keep in mind** that not all of the specs are covered, and none of the links (and buttons) in the More Panel's headers lead to anywhere (this can be changed to go to certain web sites as mentioned above). I also **recommend** this to be used in the current 10x3 grid and to have at-least 8 buttons in the main QSS panel.


amb26:
ok to test

amb26:
@Karadaliev - we would appreciate any help with understanding the current CI hangs, which are affecting pulls such as https://github.com/GPII/gpii-app/pull/178 and https://github.com/GPII/gpii-app/pull/182 - we have not had a successful CI run for gpii-app for nearly 2 weeks

Karadaliev:
Ready for review (and possible a merge?!) 🛩️

javihernandez:
@danailbd can you update it with current master?

Karadaliev:
Done, merged with master. Ready for review and/or merge @amb26 🛩️

PR #188, opened 2020-03-23
GPII-4409: MYOB launchers activate an existing application instance.

stegru:
This the MYOB application launch buttons check if there's already a running instance of the application, then activate the window. If there isn't an instance, it will be started like normal.

To test:
- Click the `Launch Notepad` button a few times.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
         "shelljs": "0.8.2"
     },
     "optionalDependencies": {
-        "gpii-windows": "0.3.0-dev.20200316T130633Z.d2a0e074"
+        "gpii-windows": "stegru/windows#GPII-4409"
     },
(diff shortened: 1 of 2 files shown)


stegru:
ok to test

unrelated fail?

```
08:10:07.879:  Test case listener has not responded after 5000ms - at sequence pos 44 of 257 sequence element {

    "changeEvent": "{that}.app.qssWrapper.qssWidget.applier.modelChanged",

    "path": "isShown",

    "listener": "jqUnit.assertTrue",

    "args": [

        "The QSS widget is shown when the QSS button is activated using spacebar",

        "{that}.app.qssWrapper.qssWidget.model.isShown"

    ]

} of fixture QSS Widget integration tests

```

PR #190, opened 2020-04-09
GPII-4424: Add siteConfig option to remove the QSS's close (X) button

Karadaliev:
Added new option inside siteConfig.json > qss > showQssCloseButton which determines if the Close (X) button of the QSS itself will be there or not.


amb26:
ok to test

PR #194, opened 2020-04-28
GPII-4422: Use Chrome for MYOB

Karadaliev:
Forcing the custom URL buttons to always use Chrome (if available).

### DIFF
--- a/src/renderer/qss/js/qssServiceButtons.js
+++ b/src/renderer/qss/js/qssServiceButtons.js
@@ -225,7 +225,9 @@
                 funcName: "gpii.windows.openUrl",
                 args: [
                     "{that}.model.item.schema.url", // using the url from the custom button's schema
-                    "{gpii.qss}.options.siteConfig.alwaysUseChrome", // Override the OS default browser.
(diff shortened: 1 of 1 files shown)


javihernandez:
ok to test

src/renderer/qss/js/qssServiceButtons.js
```diff
@@ -225,7 +225,9 @@
                 funcName: "gpii.windows.openUrl",
                 args: [
                     "{that}.model.item.schema.url", // using the url from the custom button's schema
-                    "{gpii.qss}.options.siteConfig.alwaysUseChrome", // Override the OS default browser.
+                    true, // Override the OS default browser to always use Chrome instead
+                    // if you want to use the value from site config this is:
```
amb26, line 229: Please add a link to https://issues.gpii.net/browse/GPII-4422 where this reasoning is explained

Karadaliev: Done, added to the comment just below the variable to explain the reasoning.

PR #195, opened 2020-05-07
GPII-4474: Fix the footerTip issue (showing on every menuWidget)

Karadaliev:
This fixes the small bug when you open the Color Vision first and then any menuWidget type button (Languages, High Contrast, etc.) and the color vision footerTip still shows.


Karadaliev:
@amb26 This is extremely small fix, for a bug introduced with the Color Vision PR (which was fixed manually in the previous installers). Please look at it 🛩️

javihernandez:
ok to test

PR #202, opened 2020-05-27
GPII-4410: Opening BYOB web links in an existing tab.

stegru:
This turns links in the BYOB into a special url which a browser extension (https://github.com/GPII/gpii-open-same-tab-extension) will activate a tab if the link is already loaded. Otherwise, open a new tab like normal.

The urls will look like `http://opensametab.morphic.org/redirect/https%3A%2F%2Fexample.com`.

For times when the extension isn't installed (which can't easily be detected), things should still work. So, `opensametab.morphic.org` actually points to `127.0.0.1` so Morphic can listen for the request (with a `/redirect` handler) and perform the redirect. Behind the scenes, the url will be modified so it accesses the correct port.

Meaning, when an unconfigured browser tries to access the above url, the browser will still be redirected to `https://example.com`.

Requires https://github.com/GPII/gpii-open-same-tab-extension/pull/2

To test: click on the `reddit` button - it should work with or without the browser extension.


src/main/linkRedirect.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Handles /redirect, which redirects the browser to the provided url.
+ * This is a fall-back for the 'open link in existing tab' feature [GPII-4410], where the browser is opened with a
+ * special link which gets intercepted by a browser extension. The link really points to this handler, so it will
+ * still work with browsers that don't have the extension installed.
+ */
+fluid.defaults("gpii.app.linkRedirect", {
+    gradeNames: ["kettle.app"],
+    defaultRedirectUrl: "http://opensametab.morphic.org/redirect/$1",
```
jobara, line 35: I don't see where the "refreshsametab.morphic.org" urls are redirected. Is that in configuration somewhere?

stegru: Yes, it's sort of in config.

See: https://github.com/GPII/gpii-app/pull/202/files#diff-713841a004bb552ae227e855846d4b04R34-R39

PR #203, opened 2020-06-01
GPII-4049: Volume Control v3

Karadaliev:
Moving from a settings based Volume/Mute control, to one that uses Windows real-time commands. With that we no-longer trying to read the current volume, and don't track it internally as well. Just sending volume up/down and mute commands on key presses.

**Notes:**
* As we don't have the functionality to read Windows's mute state we don't know it and assume that it's not muted. So if Windows's muted and you open QSS looks like it's not, so when you click the "Mute" button it will unmute Windows, but will show that it's muted in QSS (toggle on, and the button will show that it's muted). So the buttons will work properly, but the UX will be reversed. Gregg knows about this, and accepted it in order to finish this quicker.
* If you click on the "Mute" button this will toggle Windows's mute, and show the Mute state in QSS (the button will be red, and button's text will change). If you click Mute again this will reverse, the same goes if you click Volume Up as well, so the Volume Up will act as Unmute too.

### DIFF
--- a/src/main/common/utils.js
+++ b/src/main/common/utils.js
@@ -496,6 +496,38 @@ gpii.app.getVolumeValue = function (browserWindow, messageChannel) {
 };
 
 /**
+ * A standard volume control, it can simulate volume up, down and mute
+ * @param {String} command - accepts "up", "down" and "mute" as commands
+ */
+gpii.app.volumeControl = function (command) {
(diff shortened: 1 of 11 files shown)


Karadaliev:
@amb26 @javihernandez This is ready for testing and review! 🛩️

javihernandez:
ok to test

PR #205, opened 2020-06-03
GPII-4494 Intelligently listen to the Quickstrip's "onSaved" events

sgithens:
The capture tool is listening to the Quickstrip's "onSaved" events

### DIFF
--- a/src/renderer/captureTool/js/capture.js
+++ b/src/renderer/captureTool/js/capture.js
@@ -146,7 +146,10 @@
             // Prefs Set Name to Save the captured settings to. This is either
             // what is typed in the input for saving to a new set, or the `selectedPrefsSetName`
             // if an existing set is chosen.
-            prefsSetName: "MyCapture"
+            prefsSetName: "MyCapture",
+
+            // Value to keep track of if we are currently saving.
+            currentlySaving: false
 
         },
         modelListeners: {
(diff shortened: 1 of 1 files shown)


src/renderer/captureTool/js/capture.js
```diff
@@ -644,11 +642,13 @@
                 that.applier.change("prefsSetName", that.model.selectedPrefsSetName);
             }
             // that.render("5_confirmation");
+            that.applier.change("currentlySaving", true);
             that.saveCapturedPreferences();
         }
         else if (currentPage === "4_save_name") {
             // that.applier.change("currentPage", "5_confirmation");
```
amb26, line 649: Probably a good opportunity to clean up these comments or at least explain why they are here

src/renderer/captureTool/js/capture.js
```diff
@@ -146,7 +146,20 @@
...
-            prefsSetName: "MyCapture"
+            prefsSetName: "MyCapture",
+
+            // Value to keep track of if we are currently saving the capture preferences.
+            //
+            // Because we depend on the same model events as the rest of the QSS (preferencesSavedSuccess, and
+            // preferencesSavedError) to determine if preferences are successfully saved, we set the value of
+            // this model property to determine if we are between the second to last and final screen of the capture
+            // tool. Saving occurs between this final dialog transition. If this value is false we can safely ignore the
+            // event. If we don't track capture saving state we'll jump to the capture saved confirmation screen if
```
amb26, line 157: A better approach would instantiate a dynamic "capture" component holding the event listener only for the duration of the interaction, rather than having the capture component present all the time. I won't expect this for this review, this is just a note for future "best practices" : P

src/renderer/captureTool/js/capture.js
sgithens: Swept up!

src/renderer/captureTool/js/capture.js
sgithens: Thanks for this edifying glimpse into... THE FUTURE!

sgithens:
@amb26 I believe this is cleaned up, if you could have another look

PR #210, opened 2020-06-23
GPII-4489 Site Config Option to enable/disable capture tool.

sgithens:
Adding optional menu item generation to be able to selectively include Capture Tool

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -406,8 +406,8 @@ fluid.defaults("gpii.app.menu", {
             target: "showCaptureTool",
             singleTransform: {
                 type: "fluid.transforms.free",
-                func: "gpii.app.menu.getSimpleMenuItem",
-                args: ["Capture Tool", "onCaptureTool"]
+                func: "gpii.app.menu.getOptionalMenuItem",
+                args: ["{siteConfigurationHandler}.options.siteConfig.hideCaptureToolButton", "Capture Tool", "onCaptureTool"]
             },
             forward: {
                 excludeSource: "init"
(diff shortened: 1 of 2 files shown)


src/main/menu.js
```diff
@@ -563,6 +563,30 @@ gpii.app.menu.getSimpleMenuItem = function (label, event, payload) {
...
+/**
+  * Generates an object that represents a selectable menu item. Similar to getSimpleMenuItem,
+  * but respects an initial boolean which could be a setting from the siteConfig determining
+  * whether to display this item or not.
+  * @param {Boolean} hideItem - If true, this menu item will not be displayed. (ie. this function
+  * will not return the usual menu structure).
+  * @param {String} label - The label of the item.
+  * @param {String} event - The event to be triggered on click.
+  * @param {Object} [payload] - The payload that is to be supplied with the on click event.
+  * @return {ElectronMenuItem} A simple selectable Electron menu item.
```
amb26, line 575: Declare that the return is optional - e.g. {ElectronMenuItem|Undefined}

src/main/menu.js
```diff
@@ -563,6 +563,30 @@ gpii.app.menu.getSimpleMenuItem = function (label, event, payload) {
...
+  */
+gpii.app.menu.getOptionalMenuItem = function (hideItem, label, event, payload) {
+    if (hideItem) {
```
amb26, line 578: I'm not a religious fanatic, but single (and explicit) return-ism is slightly preferable - how about 
````
return hideItem ? undefined : {
            label: label,
            click: event,
            args: payload || {}
        };
````

